PR #39, opened 2017-02-25
Add https support

msholty-fd:
I've tested this to allow joins and messages and both seem to work. When connecting, you must use `wss` instead of `ws`.

I'd like some direction on ways to document how https is supported, so please let me know what form you'd like this in.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "@types/ws": "0.0.38",
     "clock-timer.js": "^1.1.0",
     "fossil-delta": "^0.2.5",
+    "https": "^1.0.0",
     "msgpack-lite": "^0.1.26",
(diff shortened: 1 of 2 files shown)


endel:
Hey @msholty-fd! Thank you so much for your pull-request!

I've made this change differently, though - reusing the `IServerOptions` from the `@types/uws` package, which already accepts `https.Server` as an argument. The `https` package remains as an optional dependency. 

Please let me know if this fits your needs? Thanks!

PS: I've seen you started a pretty cool project using Colyseus, that's awesome :)

msholty-fd:
@endel Thanks for the quick response!!

I checked out the most recent version and tested it out. It indeed works for me, without the need for adding a dependency to `https`. Awesome!

PR #46, opened 2017-03-29
Make version bump to support new version of uws

Ughuuu:
I cannot find a workaround to install colyseus in it's current state. After a version bump, it works, the tests pass also.


endel:
Thank you so much @Ughuuu! 🎉  🎉  🎉

PR #54, opened 2017-05-29
create a new property `autoDispose`

seiyria:
if set to true, the Room will dispose of itself when no clients are connected. if set to false, the dispose will be prevented, making the room persistent.


endel:
Awesome, simple & powerful, thanks 🎉

seiyria:
thank you!

PR #77, opened 2017-08-30
Uses the server's ephemeral port instead of the default one.

Federkun:
Fix https://github.com/gamestdio/colyseus/issues/76


endel:
Awesome! Thanks a lot, @Federkun!

Just released version `0.5.2` using your patch.

PR #79, opened 2017-08-30
Call Room#onDispose when room is discarded 

Federkun:
See https://github.com/gamestdio/colyseus/issues/78


endel:
Thanks a lot! I've made some changes in your pull-request!

PR #81, opened 2017-08-30
Fix race condition

Federkun:
Sometimes a type error occour: 

```
/src/colyseus-examples/node_modules/colyseus/lib/cluster/Worker.js:71
            var request = new http.ClientRequest({ port: server.address().port });
                                                                         ^

TypeError: Cannot read property 'port' of null
```
This should fix the race condition.

### DIFF
--- a/src/ClusterServer.ts
+++ b/src/ClusterServer.ts
@@ -109,7 +109,11 @@ export class ClusterServer {
 
     if (options.server) {
       // Don't expose internal server to the outside.
-      this.server = setupWorker(options.server.listen(0, "localhost"), this.matchMaker);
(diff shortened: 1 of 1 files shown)


endel:
Hey @Federkun, thanks for your pull-request! It seems to make sense but I'd like to be able to reproduce. 

You're using Docker, right? I've never used it, if you can share some instructions with me on how to set it up I'd be grateful :)

Thanks!

Federkun:
You should be able to reproduce it by simply send many requests when the server is starting. So, immediately after `npm start` hit many `f5`.

endel:
I see! Great job @Federkun!

PR #83, opened 2017-08-31
Fix tests that fails randomly

Federkun:
They fails because:

```
Uncaught Error: trying to broadcast null state. you should call #setState on constructor o
r during user connection.
```

So, let's provide them a state.


endel:
Great! Thanks @Federkun! <3

PR #84, opened 2017-08-31
Don't try to create a room with an invalid name

Federkun:
When an user try to join a room with an invalid name the server crash with this error:

```
/src/game/node_modules/colyseus/lib/MatchMaker.js:153
        var room = null, handler = this.handlers[roomName][0], options = this.handlers[roomName][1];
                                                          ^

TypeError: Cannot read property '0' of undefined
    at MatchMaker.create (/src/game/node_modules/colyseus/lib/MatchMaker.js:153:59)
    at MatchMaker.onJoinRoomRequest (/src/game/node_modules/colyseus/lib/MatchMaker.js:62:45)
    at process.<anonymous> (/src/game/node_modules/colyseus/lib/cluster/Worker.js:104:24)
    at emitTwo (events.js:130:20)
    at process.emit (events.js:213:7)
    at handleMessage (internal/child_process.js:753:14)
    at Pipe.channel.onread (internal/child_process.js:472:11)
worker 6586 died. Respawn.
```

This should fix that.

### DIFF
--- a/src/MatchMaker.ts
+++ b/src/MatchMaker.ts
@@ -59,8 +59,7 @@ export class MatchMaker {
 
     if (!this.hasHandler(roomToJoin) && isValidId(roomToJoin)) {
       room = this.joinById(roomToJoin, clientOptions);
-
-    } else {
+    } else if (isValidId(roomToJoin)) {
       room = this.requestToJoinRoom( roomToJoin, clientOptions ).room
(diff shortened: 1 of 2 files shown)


Federkun:
This is a break backward compatibility, sorry about that. I was half asleep yesterday. roomName !== roomId. I'll push the correct patch in a second.

PR #85, opened 2017-09-01
Dont try to create a room with invalid handler

Federkun:
This PR include:

- Correct patch for #84 + test.
- A test still fails randomly even after #83, this should finally fix that.
- Missing `stub.restore()` of a test introduced in #79.


endel:
Thanks a lot @Federkun! I've made some changes based on your pull-request.

PR #87, opened 2017-09-02
Remove express dependency

Federkun:
Right now your application require to use `express`. We could move `express` from the `package.json`'s `devDependencies` to the `dependencies`, or remove it entirely. This PR does the latter, allowing the user to do this:

```
import { createServer } from 'http';

const httpServer = createServer((req, res) => {
  // ...
});

const gameServer = new Server({ server: httpServer });
```


endel:
Nice one! Thanks again @Federkun! <3

Issue #49, opened 2017-04-14
Authentication


darkyen:
As of now Colyseus does not have any authentication step in built, so Marie can be Marie but is actually Jake! Lets make sure that Jake can't be Marie.

My proposal is as follows and assumes TLS 

- Upon initial connection the server sends a `request_authorization` event. Which has a nonce and a state.
- Upon receiving the `request_authorization` the client fetches a JWT using an IdP and sends this JWT back `authorization` as the first event with the payload as a JWT.
- The server validates the token
    - If the token is invalid or `nonce`/`state` failure occurs the server emits an `authorization_error` event and must close the connection immediately. 
    - If the token is valid the server sends `authenticated` event and continues to initialize the client. 

This will work just fine with #48 in future as any transport mechanism can implement this as the primary means of upgrading the connection. 

The Token structure is tbd we need to consider what the user will be able to do (privileged access, etc, before we find the optimal token structure)

This flow does not account for IdP initiated flow. However, the match-making can still be moved to the identity provider which will resolve #43

endel:
Maybe we can use `verifyClient` to validate client connections into rooms: https://github.com/websockets/ws/blob/master/doc/ws.md

darkyen:
The validation can happen in any step.

endel:
Finally an initial implementation of this! https://github.com/gamestdio/colyseus/wiki/Authentication

Feel free to re-open if you feel something is missing. Thanks, @darkyen!

darkyen:
LGTM

chrene:
What happened to the wiki page? @endel

endel:
@chrene the wiki has been removed long ago in favor of the new docs site https://docs.colyseus.io/server/authentication/

PR #110, opened 2017-11-11
initial auth support. #49

endel:
This is the initial implementation of #49 

This adds a new method in `Room` that you can use for handling client authentication, called `verifyClient`. You can use it synchronously or asynchronously.

The method call order is:

- `requestJoin`
- `verifyClient` < added by this pull-request
- `onJoin`

### Synchronous usage (`verifyClient(client, options): boolean`)

```typescript
import { Room } from "colyseus";

class MyRoom extends Room {
  verifyClient (client, options) {
    return (options.password === "secret");
  }
}
```

### Asynchronous usage (`verifyClient(client, options): Promise`)

By returning a promise directly:

```typescript
import { Room } from "colyseus";

class MyRoom extends Room {
  verifyClient (client, options) {
      yourDatabaseCall((err, success) => {
        if (!err) { 
          resolve();
        } else {
          reject(err);
        }
      });
    });
  }
}
```

By using `async`/`await`:

```typescript
import { Room } from "colyseus";

class MyRoom extends Room {
  async verifyClient (client, options) {
    const userData = await yourDatabaseCall();
    return (userData) ? true : false;
  }
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "uws": "8.14.1"
   },
   "devDependencies": {
-    "@types/sinon": "^2.3.7",
+    "@types/sinon": "^4.0.0",
     "assert": "^1.3.0",
     "benchmark": "^2.1.1",
     "express": "^4.16.2",
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
     "uws": "8.14.1"
   },
   "devDependencies": {
+    "@types/sinon": "^2.3.7",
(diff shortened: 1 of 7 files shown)


oyed:
I think the implementation is great - one thought on the authentication process is for (optional) authentication on initial connect. This will just prevent people from being able to connect to the Server and search for a Room whilst being unauthenticated.

Issue #120, opened 2018-01-20
Custom Server Exit Hook


oyed:
Whilst you can listen to the `SIGINT`etc. events, this will not ensure that the custom hook will execute _after_ all of the Rooms have gracefully shutdown. Having an additional, post-Room shutdown hook will better allow for state persistence.

We'd be looking at adding another Promise to the end of the room shutdown promises seen here (https://github.com/gamestdio/colyseus/blob/master/src/MatchMaker.ts#L385)

endel:
Thanks a lot @oyed!  

`Server#onShutdown` and `ClusterServer#onShutdown` are now available on version `0.8.8`.

Wiki page: https://github.com/gamestdio/colyseus/wiki/Graceful-shutdown

PR #121, opened 2018-01-21
improve graceful shutdown feature. add onShutdown() method. #120

endel:
(Solves #120)

This pull request adds `onShutdown()` method on `Server` and `ClusterServer`.

The new graceful shutdown order is:
- Match-making may dispose all rooms asynchronously
- Registered `onShutdown` sync/async method is called
- Process is killed

Usage example:

#### Using `ClusterServer`

```typescript
if (cluster.isMaster) {
  gameServer.fork();

  // shutting down master node
  gameServer.onShutdown(() => {
      setTimeout(() => {
        console.log("CUSTOM SHUTDOWN MASTER: FINISHED");
        resolve();
      }, 1000);
    })
  });

} else {
  // shutting down each worker (optional)
  gameServer.onShutdown(() => {
      setTimeout(() => {
        console.log("CUSTOM SHUTDOWN WORKER: FINISHED");
        resolve();
      }, 1000);
    })
  });
}
```

#### Using `Server`

```typescript
gameServer.onShutdown(() => {
    setTimeout(() => {
      console.log("CUSTOM SHUTDOWN: FINISHED");
      resolve();
    }, 1000);
  })
});
```

Any feedback is welcome! @oyed

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,13 +43,13 @@
     "ws": "^3.3.3"
   },
   "devDependencies": {
-    "@types/uws": "^0.13.1",
-    "@types/ip": "^0.0.29",
     "@types/debug": "^0.0.30",
-    "@types/node": "^7.0.18",
     "@types/fossil-delta": "^1.0.0",
+    "@types/ip": "^0.0.29",
     "@types/mocha": "^2.2.32",
+    "@types/node": "^7.0.18",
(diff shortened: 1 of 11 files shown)


oyed:
Looks great, exactly what I needed! Thanks 👍

PR #129, opened 2018-02-21
master

GAMEKON:
· add "experimentalDecorators" in tsconfig.json; 
· fix resolve from reject in ChatRoom; 
· add a build-in AdminRoom usage template and static admin client page like Smartfox Server 2X does.

### DIFF
--- a/static/admin.html
+++ b/static/admin.html
@@ -1,0 +1,10 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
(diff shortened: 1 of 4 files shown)


endel:
Hi @GAMEKON, thanks for your pull-request. 

Having an administration panel would be nice as an external module, not on Colyseus' core. There's one external module currently available called [colyseus-monitor](https://github.com/derwish-pro/colyseus-monitor/).

Feel free to create your own!
Cheers

PR #133, opened 2018-03-27
add @types/uws to dependencies

sandrosc:
@types/uws should be a dependency instead of a devDependency.


endel:
Hi @sandrosc,

Thanks for your time making this pull-request. 

The version `0.9.0` is going to fix this already. As `uws` will be a `peerDependency`, both `uws` and `@types/uws` should be in your own application's dependencies if you'd like to use it instead of the default `ws`.

You can already try the `0.9.0` alpha version if you'd like to. The breaking changes of `0.9.0` are:
https://github.com/colyseus/docs/blob/0.9.0/docs/migrating.md#migrating-to-09x-from-06x-or-08x (some links are broken because the docs need to be compiled into html)

Here's how to integrate with `uws` on `0.9.0`:
https://github.com/colyseus/docs/blob/0.9.0/docs/api-server.md#optionsengine

Latest alpha versions are:
- `colyseus@0.9.0-alpha.10` (server)
- `colyseus.js@0.9.0-alpha.5` (client)

Issue #138, opened 2018-04-04
Allow disabling automatic patches


endel:
There are some use cases that state sync is not necessary at all.

Currently, the following doesn't have the expected behavior:

```
this.setPatchRate(null);
this.setPatchRate(0);
```

By passing such values on `setPatchRate`, the [automatic patch broadcasts](https://github.com/gamestdio/colyseus/blob/f3f0c217eed04a0c1a44821a833954675b85971f/src/Room.ts#L231) should be disabled.

terranblake:
@endel I am looking into this issue as my first pull request, and am wondering if the following test is sufficient for the expected behavior that you are looking to reconcile in this issue?

`it('should disable "patch" interval', function () {
      var room = new DummyRoom();
      room.setPatchRate(null);

      assert.equal("object", typeof ((<any>room)._patchInterval));
      assert.equal(-1, (<any>room)._patchInterval._idleTimeout, "patch rate should be disabled, set to -1");
    });`

Where `-1` is the resultant value after using `clearTimeout()` to disable the patch timer. I believe I have a fix, and it does pass this test, but I'm not sure if I am missing something.

terranblake:
I have also tested this using the `colyseus-examples` repository, which resulted in successfully disabling the patch timer. All other tests continue to pass.

GAMEKON:
what about

use patchRate property instead of setPatchRate() method.¶
OLD


constructor (options) {
  this.setPatchRate(1000 / 50);
}
NEW


class MyRoom extends Room {
  patchRate = 1000 / 50;
}

PR #139, opened 2018-04-06
Allowed Disabling of Automatic Patches #138

terranblake:
This is a proposed solution to #138 and should correct the unexpected behavior described by @endel in the original issue. Automatic patch broadcasts will now be disabled when the following occurs.

```
this.setPatchRate(null);
this.setPatchRate(0);
```

This is my first pull request, so feel free to give any advice, or concerns on the format of this request, as well as the small bit of code that I actually ended up changing.


src/Room.ts
```diff
@@ -119,9 +119,11 @@ export abstract class Room<T= any> extends EventEmitter {
 
   public setPatchRate( milliseconds: number ): void {
     // clear previous interval in case called setPatchRate more than once
-    if ( this._patchInterval ) { clearInterval(this._patchInterval); }
-
-    this._patchInterval = setInterval( this.broadcastPatch.bind(this), milliseconds );
+    if ( milliseconds === null || milliseconds === 0 ) {
+      clearInterval( this._patchInterval );
```
endel: `clearInterval` would be always necessary to prevent multiple intervals from being registered in case `setPatchRate` is called multiple times.

endel:
That's great @terranblake! Thanks a lot for taking your time on this. I've left a comment in your code. Once that's fixed I'll merge it!

Cheers!

terranblake:
Ah! I completely overlooked that. Also, @endel I'm not sure if explaining the cause of the issue is common practice, but I'll go ahead and share the issue that I experimented with.

Essentially, the Timer object, when passed `0` repeatedly dispatches the `broadcastPatch()` callback, and, when passed `null`, it defaults to 1ms.

I have included the requested revision. Should be ready to merge!

endel:
Thanks a lot @terranblake! Just released version `0.9.3` with your patch. Cheers!

Issue #147, opened 2018-04-29
Allow to reconnect client into the room


endel:
When the WebSocket connection drops unexpectedly, it should be possible to reconnect that client within a certain timespan, even if the room is locked.

**Scenarios:**

- User closed browser tab accidentaly ([colyseus forum](https://discuss.colyseus.io/topic/71/onleave-timeout/3))
- Mobile device went to sleep mode (locked) ([defold forum](https://forum.defold.com/t/colyseus-multiplayer-game-client-server/17101/2))
- Switching from Wi-Fi networks. (#8)
- ...any unexpected action that closes the WebSocket connection. (#68)

endel:
I've set a poll to gather what you guys think about this feature being enabled by default or not. What do you think?

[![](https://api.gh-polls.com/poll/01CDFXAWAZTHJB7223G5NVM4CN/enabled%20by%20default)](https://api.gh-polls.com/poll/01CDFXAWAZTHJB7223G5NVM4CN/enabled%20by%20default/vote)
[![](https://api.gh-polls.com/poll/01CDFXAWAZTHJB7223G5NVM4CN/disabled%20by%20default)](https://api.gh-polls.com/poll/01CDFXAWAZTHJB7223G5NVM4CN/disabled%20by%20default/vote)

endel:
Hey @dossy @sandrosc @AnubisCode @ruifly2012,

I'd appreciate if you all could vote on the poll at the pull request #154 

Cheers!

dossy:
I voted in favor of on-by-default, although that _breaks backwards compatibility_ in case anyone implemented their own reconnection mechanism in their own codebase, but ultimately for future developers, I think the on-by-default will be the useful behavior.

PR #154, opened 2018-05-14
Allow client reconnection

endel:
(Implements #147)

I've set a poll to gather what you guys think about this feature being enabled by default or not. What do you think?

[](https://api.gh-polls.com/poll/01CDFXAWAZTHJB7223G5NVM4CN/enabled%20by%20default/vote)
[](https://api.gh-polls.com/poll/01CDFXAWAZTHJB7223G5NVM4CN/disabled%20by%20default/vote)

Things to consider when reconnection is enabled:
- The rooms will live for a bit longer (15 seconds by default), even if there're no more clients connected to it and `autoDispose` is set to `true`.
- Reconnecting clients will skip `requestJoin` method, since it was authorized already before.

**Usage**

```typescript
class MyRoom extends Room {
  onJoin (client) {
    this.state.setupClient(client.sessionId);
  }

  async onLeave (client) {
    this.state.inactivateClient(client.sessionId);

    try {
      // allow client to reconnect within 20 seconds
      const reconnectedClient = await this.allowReconnection(client, 20); 

      // client reconnected!
      this.state.reactivateClient(client.sessionId);

    } catch (e) {
      // client couldn't reconnect within 20 seconds
      this.state.removeClient(client.sessionId);
    }
  }
}
```

### DIFF
--- a/src/MatchMaker.ts
+++ b/src/MatchMaker.ts
@@ -42,7 +42,7 @@ export class MatchMaker {
     const auth = client.auth;
 
     // assign sessionId to socket connection.
-    client.sessionId = await this.presence.hget(roomId, client.id);
+    client.sessionId = await this.presence.get(`${roomId}:${client.id}`);
 
     // clean temporary data
     delete clientOptions.auth;
@@ -86,9 +86,6 @@ export class MatchMaker {
         this.remoteRoomCall(roomId, '_emitOnClient', [client.sessionId, 'close']);
       });
     }
-
-    // clear reserved seat of connecting client into the room
-    this.presence.hdel(roomId, client.id);
   }
 
   /**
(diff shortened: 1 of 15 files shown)


sandrosc:
Will we be able to change allowReconnection dynamically?

endel:
Hey @sandrosc, yes, it's possible to enable and disable `allowReconnection` at any time.

sandrosc:
@endel awesome!

endel:
It turns out I've overlooked this feature, and the implementation will need to be slightly different. I've updated the usage example in the description of this pull-request.

You may call `allowReconnection()` per client, and it will return a promise, that is resolved whenever the user has reconnected. The promise is rejected after the timeout specified. (15 seconds by default)

This way you have full control of the events, and you can flag users as inactive accordingly.

Wenish:
if i would upgrade to the version. do i need to change something so that everything works the same?

endel:
@Wenish `onAuth` will continue to be called before joining. I think `requestJoin` won't be necessary since this method is used to determine which room the client will join, and we already know exactly which room the client should connect to. 

If you're using version `0.9.x` there's nothing else to update! `allowReconnection` is totally optional.

PR #157, opened 2018-05-24
update issue template

Wenish:
it help to create a new issue.
and the issue gets more context than just some phrases..

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -3,7 +3,7 @@ Looking for help? Post on the forum instead: http://discuss.colyseus.io
 -->
 
 Your issue may already be reported!
-Please search on the [issue track](../) before creating one.
+Please search on the [issue tracker](../) before creating one.
 
 ## Steps to Reproduce (for bugs)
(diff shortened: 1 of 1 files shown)


endel:
Hey @Wenish, sorry the delay to reply here.

I believe requiring `Expected Behavior`, `Current Behavior` and `Possible Solution` might be too much, since most users of Colyseus have little to no knowledge of Node.js or networking. I like having `Context` and `Your Environment`, though. 

Can you make this change? Cheers!

Wenish:
@endel np ;)

i took the 3 things out and added some more example lines in the "your environment" section.

seiyria:
Can you change "issue track" to "issue tracker"? Additionally, I think you should remove unity version from this template - if desired, it could be added to the unity plugin template.

oyed:
@seiyria Agreed - Client-based frameworks should be restricted to their own repos issues, not `colyseus`

Wenish:
@seiyria i removed the line with unity and browser version

Wenish:
@endel what do u think?

endel:
Thanks @Wenish, it's looking good! 🤗

PR #182, opened 2018-10-03
Fix Auto dispose room while allowReconnection is enable

bhavik66:
Room was not disposing of because of reservedSeats is not deleted after reconnection timeout when allowReconnection is enabled.

### DIFF
--- a/src/Room.ts
+++ b/src/Room.ts
@@ -344,11 +344,10 @@ export abstract class Room<T= any> extends EventEmitter {
     if (allowReconnection) {
       // store reference of the roomId this client is allowed to reconnect to.
       this.presence.setex(client.sessionId, this.roomId, seconds);
+    } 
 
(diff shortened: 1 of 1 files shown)


endel:
Hi @bhavik66, 

When `allowReconnection()` is called, there's another timeout created here to reject the reconnection, which also clears the reserved seat: https://github.com/gamestdio/colyseus/blob/master/src/Room.ts#L318

So, I'm afraid this change won't help on that case.

bhavik66:
Yes, I know `cleanup` is method for clear the reserved seat, But It's called after this.resetAutoDisposeTimeout(seconds); so when It's trying to check _disposeIfEmpty there was reserved seat is not be clear and Room is not dispose

https://github.com/gamestdio/colyseus/blob/master/src/Room.ts#L370

Issue #171, opened 2018-07-27
Async `onLeave` causes unhandled error when client disconnects


oyed:
When the Client disconnected immediately, such as refreshing the page, and the Server has an async `onLeave`, the following error is triggered:

`WebSocket is not open: readyState 2 (CLOSING)`

Presumably because Colyseus is trying to send the event stating that `onLeave` has finished, but the client is already disconnected.

oyed:
This seems to be the `send` call causing it: https://github.com/gamestdio/colyseus/blob/master/src/MatchMaker.ts#L60

endel:
Thanks for reporting @oyed, I'm having the same issue sometimes and didn't know the cause of it!

endel:
Stack trace reported from @seiyria on [gitter](https://gitter.im/gamestdio/colyseus):

```
Error: WebSocket is not open: readyState 3 (CLOSED)
1
at WebSocket.send (/app/node_modules/ws/lib/websocket.js line 314 col 19)
const err = new Error(
2
at presence.subscribe (/app/node_modules/colyseus/lib/MatchMaker.js line 46 col 32)
client.send(new Buffer(data), { binary: true });
3
at <unknown> (subscriptions.(anonymous function) (/app/node_modules/colyseus/lib/presence/MemsharedPresence.js line 17 col 50)
4
at <unknown> (/app/node_modules/memshared/lib/commands/pubsub.js line 73 col 64)
subscriptions[topic].forEach(function (c) { return c(message); });
5
at Array.forEach (<anonymous> line (unknown))
6
at publish (/app/node_modules/memshared/lib/commands/pubsub.js line 73 col 34)
subscriptions[topic].forEach(function (c) { return c(message); });
7
at process.workerHandleIncomingMessage (/app/node_modules/memshared/lib/index.js line 30 col 31)
commands[message.cmd].apply(undefined, message.args);
8
at process.emit (events.js line 187 col 15)
9
at emit (internal/child_process.js line 811 col 12)
10
at process._tickCallback (internal/process/next_tick.js line 63 col 19)
```

seiyria:
Possibly another error related to this:

```
2018-10-17T18:56:13.687660502Z app[web.1]: TypeError: First argument must be a valid error code number
2018-10-17T18:56:13.687677345Z app[web.1]:     at Sender.close (/app/node_modules/ws/lib/sender.js:112:13)
[...]

PR #189, opened 2018-10-18
only send data when client is OPEN

seiyria:
not sure specifically if this is how you'd want this implemented, but let me know and I can update it.

#171

### DIFF
--- a/src/MatchMaker.ts
+++ b/src/MatchMaker.ts
@@ -1,3 +1,5 @@
+import * as WebSocket from 'ws';
+
 import { merge } from './Utils';
 
 import { Client, generateId, isValidId } from './index';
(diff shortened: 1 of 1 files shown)


oyed:
I think this should be enough, however it doesn't stop the calls made being sent in the first place - probably something to look in to via a separate PR/issue.

seiyria:
That's true. This will prevent the errors from spreading but debugging why they happen was tougher since I can only ever reproduce these things on prod, but my build process doesn't let me adequately test colyseus changes on prod.

endel:
Thanks a lot @seiyria, there was some checks for `OPEN` connections in some places, and now I've centralized it in the `send` method, which is used everywhere. (https://github.com/gamestdio/colyseus/commit/5c37e1bed34b55769c95e4ce5d1a6918306e5000)

seiyria:
Great. That works for me, thanks!

Issue #188, opened 2018-10-18
Shutdown signal handling should happen outside of Colyseus


jmeyers91:
Colyseus should not be calling `process.exit`. I'm integrating a colyseus server in an existing app that has its own graceful shutdown code, but the colyseus graceful shutdown is cutting off my project's cleanup by killing the process before it has finished. In my opinion, colyseus should just expose a `shutdown` or `dispose` method and leave it up to the container application to decide when to run them.

gioragutt:
How about adding an option to the Colyseus server, something like `shutdownProcessOnSignal`, which is by default `true`, to retrain the existing behaviour, but can be turned to false in order to disable the behaviour you explained above?

jmeyers91:
@gioragutt That sounds like good solution. I submitted a PR.

PR #190, opened 2018-10-22
Add server option shutdownProcessOnSignal

jmeyers91:
This option changes nothing by default, but if it is set to `false` it disables the automatic process shutdown when a kill signal is received.

See #188 for reasoning.

### DIFF
--- a/src/Server.ts
+++ b/src/Server.ts
@@ -24,6 +24,7 @@ export type ServerOptions = IServerOptions & {
   presence?: any,
   engine?: any,
   ws?: any,
+  shutdownProcessOnSignal?: Boolean,
 };
 
 export class Server {
@@ -48,12 +49,16 @@ export class Server {
     // "presence" option is not used from now on
     delete options.presence;
 
-    registerGracefulShutdown((signal) => {
-      this.matchMaker.gracefullyShutdown().
-        then(() => this.shutdown()).
-        catch((err) => debugError(`error during shutdown: ${err}`)).
-        then(() => process.exit());
-    });
+    if(options.shutdownProcessOnSignal !== false) {
+      registerGracefulShutdown((signal) => {
+        this.matchMaker.gracefullyShutdown().
+          then(() => this.shutdown()).
+          catch((err) => debugError(`error during shutdown: ${err}`)).
+          then(() => {
+            process.exit()
+          });
+      });
+    }
 
     if (options.server) {
       this.attach(options);
--- a/src/Server.ts
+++ b/src/Server.ts
@@ -24,7 +24,7 @@ export type ServerOptions = IServerOptions & {
   presence?: any,
   engine?: any,
   ws?: any,
-  shutdownProcessOnSignal?: Boolean,
+  shutdownProcessOnSignal?: boolean,
 };
 
 export class Server {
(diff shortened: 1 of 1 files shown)


endel:
Hi @jmeyers91, thanks for your pull-request! Have you managed to integrate Colyseus' `matchMaker. gracefullyShutdown()` method in your app? It would be great to have an example on how to achieve this manually, too.

Cheers!

src/Server.ts
```diff
@@ -48,12 +49,16 @@ export class Server {
     // "presence" option is not used from now on
     delete options.presence;
 
-    registerGracefulShutdown((signal) => {
-      this.matchMaker.gracefullyShutdown().
-        then(() => this.shutdown()).
-        catch((err) => debugError(`error during shutdown: ${err}`)).
-        then(() => process.exit());
-    });
+    if(options.shutdownProcessOnSignal !== false) {
```
gioragutt: I think that just checking for truthiness looks a bit clean

```suggestion
    if(options.shutdownProcessOnSignal) {
```

jmeyers91: The `!== false` keeps this from breaking existing behavior since users will need to pass `shutdownProcessOnSignal: false` explicitly to disable the shutdown process. Without it, everyone who updates will need to pass `shutdownProcessOnSignal: true` to keep the existing behavior.

gioragutt: So this is kinda weird, because basically you're relying on the fact that only `false` is the disabling value, whereas `undefined` and `null` are values that enable the behaviour.

Considering the name of the flag, and the fact that it's enabled by default, there should be a place which sets a default value of `true` to the flag (maybe in the wrapping function parameter destructuring), and this will mean that by default, you need to check for a truthy (`!!`) value to turn on the behaviour.

jmeyers91: Yeah that is a bit strange. I changed it to use a default value and destructuring:

```
const { shutdownProcessOnSignal = true } = options;
if(shutdownProcessOnSignal) {
```

gioragutt: Looks better 👍 gj

src/Server.ts
```diff
@@ -24,6 +24,7 @@ export type ServerOptions = IServerOptions & {
   presence?: any,
   engine?: any,
   ws?: any,
+  shutdownProcessOnSignal?: Boolean,
```
gioragutt: From `tslint`:

```suggestion
  shutdownProcessOnSignal?: boolean,
```

src/Server.ts
```diff
@@ -48,12 +49,17 @@ export class Server {
...
     delete options.presence;
 
-    registerGracefulShutdown((signal) => {
-      this.matchMaker.gracefullyShutdown().
-        then(() => this.shutdown()).
-        catch((err) => debugError(`error during shutdown: ${err}`)).
-        then(() => process.exit());
-    });
+    const { shutdownProcessOnSignal = true } = options;
+    if(shutdownProcessOnSignal) {
```
gioragutt: From `tslint`:

```suggestion
    if (shutdownProcessOnSignal) {
```

src/Server.ts
```diff
@@ -48,12 +49,17 @@ export class Server {
...
+      registerGracefulShutdown((signal) => {
+        this.matchMaker.gracefullyShutdown().
+          then(() => this.shutdown()).
+          catch((err) => debugError(`error during shutdown: ${err}`)).
+          then(() => {
+            process.exit()
```
gioragutt: From `tslint`:

```suggestion
            process.exit();
```

endel:
Thanks a lot @jmeyers91 for the pull-request, and @gioragutt for reviewing it. I've refactored your solution quite a bit and it's now available on version [`0.9.18`, plase check the release notes](https://github.com/gamestdio/colyseus/releases/tag/0.9.18)!

PR #194, opened 2018-11-10
Allow client to force create a room on join

gioragutt:
Basically, I took inspiration from the unity3d `ColyseusClient` which passes a `create` prop to the client options.

I renamed it to `forceCreate` to make it clear that it's going to create a room and not join an available one.

This doesn't affect the `rejoin` mechanism, since this is only done when the used it joining a room by the room name, and not room id, and he's not rejoining.

### DIFF
--- a/src/MatchMaker.ts
+++ b/src/MatchMaker.ts
@@ -1,4 +1,3 @@
-import * as WebSocket from 'ws';
 import { merge } from './Utils';
 
 import { Client, generateId, isValidId } from './index';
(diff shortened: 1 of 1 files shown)


gioragutt:
I tried publishing my own fork, but kinda failed (due to everything still requiring `colyseus` as the package name`).

I went in the node_modules of my server and changed the code in `MatchMaker.js`, and this of course worked.

I would love if you guys could make a quick patch release on that, so that I can update my server, and not have to manually change `node_modules` 😅

gioragutt:
The reason I did that was to align the capabilities of the framework with those of what games actually have today. Maybe my naive implementation is easy to take advantage of, but considering you don’t have an authentication mechanism, disallowing that feature from certain users isn’t really possible in the framework, and is up to the users of the library to implement themselves. In that case, this flag works fine, so I believe. 

If you have any thoughts on how this can be improved without an authentication mechanism, I’d love to know.

I’ll be honest tho, I did stop using the framework. It’s mainly because the unity connector is so hard to work with. I think that without shitcoding your codebase to deal with the patches mechanism, you won’t be have a way to deal with a lot of data, in terms of types and complexity of the data, not the quantity of data transferred.

gioragutt:
@endel any thoughts on this?

endel:
Closing as this is not going to be implemented as part of the framework, you may always implement this in your own room handler, as described in [this example](https://github.com/colyseus/colyseus-examples/blob/ac093fa9626b5b72dcfda40e564862e6777f5476/rooms/04-create-or-join-room.ts#L14-L18).

PR #198, opened 2018-11-19
Fix async onDispose during graceful shutdown

LightningK0ala:
## Problem

There is a bug during graceful shutdown. If onDispose returns a promise it won't wait for it to resolve before the process terminates. This bug doesn't happen when a room is being automatically disposed if clients have left and autoDispose property is set to true.

## Steps to reproduce

In a room handler, setup the onDispose method to something like this:
```
onDispose() {
    console.log("onDispose: Timeout started");
    setTimeout(() => {
      console.log("onDispose: Timeout complete");
      resolve();
    }, 2000);
  });
}
```
Then connect clients to a room and terminate the running process (CTRL+C on the terminal)
You should see the first message `onDispose: Timeout started` printed but not the second.

## Cause

In Room.ts the constructor sets up a listener for "dispose" to handle proper disposing of the room (and eventually running `this.onDispose` if set). During graceful shutdown, the `disconnect()` method also registers another listener for "dispose" which simply resolves it without waiting for the other listener to complete any async calls.

## Solution

My proposed solution is to emit a "disposeComplete" event when the main "dispose" listener (in  Room.ts constructor) completes, and have the disconnect method register a listener for "disposeComplete" instead.

### DIFF
--- a/src/Room.ts
+++ b/src/Room.ts
@@ -84,7 +84,10 @@ export abstract class Room<T= any> extends EventEmitter {
 
     this.presence = presence;
 
-    this.once('dispose', () => this._dispose());
+    this.once('dispose', async () => {
+      await this._dispose();
+      this.emit('disposeComplete');
+    });
 
     this.setPatchRate(this.patchRate);
   }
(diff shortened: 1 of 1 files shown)


endel:
Nice catch @LightningK0ala! Thanks a lot for this. 

(Not sure why GitHub didn't recognize this as merged, but here's your commit in the history ❤️  https://github.com/gamestdio/colyseus/commit/e8c67f28c7ff76614f54cd7eb0a42965cebe8785)

Issue #231, opened 2019-04-15
Documentation is lacking basic info for v0.10


Mykybo:
Trying to implement the new schema introduced by v0.10 I was stuck on error: `SchemaSerializer error. See: https://docs.colyseus.io/migrating/0.10/#new-default-serializer`. 

The thing is I was using the new schema so I started debugging.

It turns out I was missing a room annotation: `@serialize(SchemaSerializer)` which isn't mentioned anywhere in docs that you have to use. 

The only context this annotation is mentioned is with fossil serializer.

endel:
hey @Mykybo, that's odd, because the `SchemaSerializer` is used by default, it shouldn't be necessary to use `@serialize(SchemaSerializer)`. How does your room handler and schema look like? Cheers

Mykybo:
After playing around with this a bit more I am very confused... 

What I did to rule out any mistakes in my implementation of the colyseus server is copy the example server from https://github.com/colyseus/colyseus/tree/master/usage, replace the relative imports with imports from the npm package and run it.

btw, you are using the serialize decorator here:
```
@serialize(SchemaSerializer)
export class ChatRoom extends Room<State> {
```

When I run this server with the decorator, I get the following error: 
```
/src/game-server/node_modules/colyseus/lib/serializer/Serializer.js:5
        constructor.prototype._getSerializer = () => new serializer();
                                             ^

TypeError: Cannot set property '_getSerializer' of undefined
    at /mnt/c/Users/username/repository/path/node_modules/colyseus/lib/serializer/Serializer.js:5:46
    at Object.decorateConstructor (/mnt/c/Users/username/repository/path/dist/server.js:1819:3578)
    at Object.decorateClass (/mnt/c/Users/username/repository/path/dist/server.js:1819:2019)
    at _decorate (/mnt/c/Users/username/repository/path/dist/server.js:1817:308)
    at Module../src/ChatRoom.ts (/mnt/c/Users/username/repository/path/dist/server.js:1865:16)
    at __webpack_require__ (/mnt/c/Users/username/repository/path/dist/server.js:20:30)
    at Module../src/server.ts (/mnt/c/Users/username/repository/path/dist/server.js:1995:67)
    at __webpack_require__ (/mnt/c/Users/username/repository/path/dist/server.js:20:30)
    at ./node_modules/@colyseus/schema/lib/ChangeTree.js.Object.defineProperty.value (/mnt/c/Users/username/repository/path/dist/server.js:84:18)
[...]

PR #233, opened 2019-04-17
Fix schema type checking

Mykybo:
Fixes: https://github.com/colyseus/colyseus/issues/231

The problem is that `instanceof` does not work when the origin of the checked class is different then the one the instance was created with.

Using constructor names to replace `instanceof` is reliable as long as the name is unique enough (to avoid naming colisions), which I think `Schema` is.

### DIFF
--- a/src/Utils.ts
+++ b/src/Utils.ts
@@ -117,3 +117,13 @@ export function logError(err: Error): void {
     debugAndPrintError(`websocket error: ${err.message}\n${err.stack}`);
   }
 }
+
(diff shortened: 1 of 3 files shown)


endel:
Thanks for taking the time for this, but this problem can be avoided when using this configuration: https://github.com/endel/coly-test/commit/c45d9d1ce6ce16b7a5db7b16f4f88ac6c599ce8e (more info https://github.com/colyseus/colyseus/issues/231#issuecomment-484304086)

Mykybo:
I see, thanks for the fix! :)

PR #234, opened 2019-04-19
fix: typing of setState argument

ChristianIvicevic:
There was an issue wherein the created typing for the setState method would accept any state instead of T. This commit should fix this.


endel:
Thanks for noticing that! just published `0.10.6` with this!

PR #238, opened 2019-05-05
Fix for rejoining with room name.

chsqn:


endel:
Closing this as matchmaking has been completely changed on version `0.11`. See migration guide here: https://docs.colyseus.io/migrating/0.11/

PR #254, opened 2019-08-23
Server add property pingCountMax

GDdark:
Set pingCountMax can let some unstable users do not frequently drop the line


endel:
Hi @GDdark, interesting, thanks for this. Can you make the pull-request against the `master` branch instead of `0.10`? Cheers!

PR #255, opened 2019-08-24
Server add property pingCountMax

GDdark:
Set pingCountMax can let some unstable users do not frequently drop the line


endel:
Looking good, thanks a lot @GDdark ❤️

PR #265, opened 2019-09-14
Move onJoin call after connection resolved

### LINKED ISSUE #260: Calling `broadcast()` during `onJoin()` is ignoring the new client
Reported by @damian-pastorini

giraffesyo:
This resolves #260 

I moved the `onJoin` call to be after all other connection logic. I tested this to see if it works and broadcast now sends to the client that just connected and all other clients from within `onJoin`. Also, `clients.length` now returns the correct number from within the `onJoin` function.

If someone was relying on the length number being wrong then this would be a breaking change for them.

### DIFF
--- a/src/Room.ts
+++ b/src/Room.ts
@@ -321,9 +321,6 @@ export abstract class Room<T= any> extends EventEmitter {
           throw new Error('onAuth failed.');
         }
 
-        if (this.onJoin) {
-          await this.onJoin(client, options, client.auth);
-        }
       } catch (e) {
         debugAndPrintError(e);
         throw e;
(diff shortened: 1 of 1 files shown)


endel:
Hi @giraffesyo, thanks for taking the time to dive into this.

I understand issue #260 is very annoying. I think another approach would be necessary actually, because when this piece of code gets called, the `client.joinOrCreate()` promise will be fulfilled:

https://github.com/colyseus/colyseus/blob/433021bd2a3732238dbe866d9566b9f7a7fda8c0/src/Room.ts#L342-L347

This means that you couldn't throw errors during `onJoin()` to reject the client-side promises (`client.joinOrCreate()`, `client.join()`, etc).

On second thought, this may be a fine breaking change. Users would be able to reject promises during `onAuth()`, but not during `onJoin()`. May not be so intuitive though.

giraffesyo:
Hi @endel, thanks for your reply!

What if these are split into two lifecycle hooks like `willJoin` and `didJoin`? I think this might be more intuitive.

endel:
Thanks for your PR @giraffesyo, version `0.12` has fixed #260 in a slightly different way! I don't want to introduce any major breaking change in the API's (e.g. `didJoin` / `willJoin`) until we reach `1.0`. Cheers!

PR #283, opened 2019-11-17
Allow to limit the number of rooms a process can spawn closes #155

### LINKED ISSUE #155: Allow to limit the number of rooms a process can spawn
To prevent the process from consuming too much CPU and slowing down all of its spawned rooms, it would be best to have a way to limit the number of rooms a process can spawn.

(https://discuss.colyseus.io/topic/11/about-setting-number-of-rooms-for-a-registered-room-on-the-server/4)

MateusMendesSantana:
Hi,

I added an instanceCount variable in RegisteredHandler to store the number of instances created for a given room. Then I could check if a new room exceeds the maximum instance number defined in instanceLimit.

### DIFF
--- a/src/MatchMaker.ts
+++ b/src/MatchMaker.ts
@@ -234,6 +234,13 @@ export async function createRoom(roomName: string, clientOptions: ClientOptions)
     throw new MatchMakeError(`"${roomName}" not defined`, Protocol.ERR_MATCHMAKE_NO_HANDLER);
   }
 
+  if (!registeredHandler.isAvailableInstantiation()) {
+    throw new MatchMakeError(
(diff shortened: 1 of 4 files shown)


endel:
Hi @MateusMendesSantana, thanks for your PR, this is a valuable feature, but I just want to add a few notes here, as of https://github.com/colyseus/colyseus/pull/309 landing:

- https://github.com/colyseus/colyseus/pull/309 has a very simple structure counting the total amount of rooms the process is handling - we can adapt the existing code to restrict based on specific types of rooms, instead of all of them
- Ideally, other processes should be able to handle spawning the requested room, in case the current process has exceeded the limit allowed
- What happens if any of the active processes can handle the "create room" request?

PR #289, opened 2020-01-03
[0.12.x] Unit tests for SchemaSerializer (recursiveness)

Zielak:
I've narrowed your test down to simply "try navigating in recursive structure".

And second failing test as a bonus, which I also found in my game. The issue may lie in here:
```
// SchemaSerializer.ts:94
const childSchema = (schema[fieldName][0] as typeof Schema)._schema;
```
For a map of strings, its trying to get `_schema` property of primitive value (eg. string) -> undefined, and inside the next `this.hasFilter(undefined, undefined)` it breaks.

+ git ignore coverage reports
+ easier tests debugging in vscode

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,6 @@ node_modules
 yarn.lock
 /lib
 .DS_Store
+
+.nyc_output
+coverage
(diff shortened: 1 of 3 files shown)


endel:
Thanks a lot @Zielak, I'll continue from there! ❤️

PR #290, opened 2020-01-04
[0.12.x] More recursive schema

Zielak:
Simply checking if parent is the same as new schema wasn't enough in my case, so my proposal is to add a Set of all schemas the function touched so far.

The server-side of my app now seems to work fine (no "Maximum call stack size exceeded" error), but my client-side doesn't react on client joining. I'm already linked with colyseus.js#0.12.x branch. So I actually can't fully confirm the fix.

I've done 2 commits, one with the fix+tests and other with "housekeeping" things (added package-lock, tslint for tests). You can keep only the one with the fix if that other one was too much.

Cheers.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,7562 @@
+{
+  "name": "colyseus",
+  "version": "0.12.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.5.5",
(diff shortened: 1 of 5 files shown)


endel:
Awesome @Zielak, thanks! 🎉  I'm starting to publish `colyseus@alpha` and `colyseus.js@alpha` for 0.12.x, you mind trying them out?

Zielak:
I just installed 0.12.x alpha for both server and client, and it seems to work fine! Client joins the room, state and interactions work! 🥇

PR #309, opened 2020-02-01
Spawn rooms on remote processes

endel:
> I'm going to test this on production soon. Leaving this as a draft until it actually works for my use-case!

This PR includes a few housekeeping changes and the ability for any process to ask for other processes to create the room, whenever it's requested.

- When the process is initialized Presence receives `roomcaches[processId] = 0`
- Whenever a room is created, Presence is updated to `roomcaches[processId]++`
- Whenever a room is removed, Presence is updated to `roomcaches[processId]--`
- During shutdown, Presence is updated to `delete roomcaches[processId]`

When a "create room" request is received on any process, the `roomcaches` is queried, and the process with the least amount of rooms will be used to create the room, via inter-process communication (`IPC.ts`)

The same applies when using `matchMaker.createRoom()` from the backend.

## Coming soon?

I believe this approach can evolve to support having multiple machines with a fixed ENDPOINT for each of them and having a regular load balancer behind all the machines. So when requesting to matchmake, the server can respond with the ENDPOINT + roomId, and the client could then connect directly to that endpoint, instead of relying on a single endpoint, as it works right now with the proxy.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "colyseus",
-  "version": "0.12.0",
+  "version": "0.12.1-alpha.0",
   "description": "Multiplayer Game Server for Node.js.",
   "main": "./lib/index.js",
   "typings": "./lib/index.d.ts",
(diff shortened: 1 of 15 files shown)


endel:
This update has been running for a few days, and it seems pretty stable!

CPU is being spread across the two active processes, and in the `hgetall roomcount` command, you can see the number of rooms each process is currently handling.

Screenshot 2020-02-03 at 14 16 43

PR #314, opened 2020-02-21
Fix disconnect().

chsqn:
clients[] don't contain clients of reservedSeats[]/reconnections[].

### DIFF
--- a/src/Room.ts
+++ b/src/Room.ts
@@ -291,6 +291,10 @@ export abstract class Room<State= any, Metadata= any> extends EventEmitter {
     const delayedDisconnection = new Promise((resolve) =>
       this.once('disconnect', () => resolve()));
 
+    for (const reconnection of Object.values(this.reconnections)) {
+      reconnection.reject();
+    }
+
     let numClients = this.clients.length;
     if (numClients > 0) {
       // prevent new clients to join while this room is disconnecting.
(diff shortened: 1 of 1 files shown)


endel:
Nice catch @chsqn, thanks for this. I've just released `0.12.2` with this. Cheers! ❤️

grant1842:
Hmm I wonder if this is what has been driving me crazy with handling multiple clients with the same access token

chsqn:
For your reference.

1. autoDispose will change to true:
https://github.com/colyseus/colyseus/blob/e2c0aceaa8e6ad32018435a224ec661e4fed6750/src/Room.ts#L287-L289

2. onDispose() is invoked before onLeave() completes in disconnect()
Therefore I need manage when/where to save/remove players.
```
  async onLeave (client, consented) {
    const player = this.state.players[client.sessionId]
    var isRemoval
    if (consented) {
      isRemoval = !this.autoDispose
    } else {
      isRemoval = await this.allowReconnection(client).then(function () {
        return false
      }).catch(function () {
        if (this.autoDispose) {
          return false
        } else {
          return true
        }
      })
    }
    if (isRemoval) {
      await player.save()
      delete this.state.players[client.ticketStub]
      if (Object.keys(this.state.players).length === 0) {
        this.disconnect()
      }
    }
  }
  async onDispose () {
    for (const player of Object.values(this.state.players)) {
      await player.save()
    }
  }
```

PR #327, opened 2020-04-24
Updates discovery to allow for Environment Variable overrides

LukeWood:
By allowing environment variables to override the ip address and port
number used in the discovery api we can support in cluster routing (i.e.
nodes that do not have access to the external net).

An example of this would be routing to Game Servers contained behind a
kubernetes service
(https://kubernetes.io/docs/concepts/services-networking/service/).
Services by default do not have an external IP address assigned to them:
only an internal cluster IP.

When creating our deployments we can specify OVERRIDE_IP_ADDRESS and
OVERRIDE_PORT environment
variables.
These will allow internal cluster routing to work with the
existing proxy system.

### DIFF
--- a/src/discovery/index.ts
+++ b/src/discovery/index.ts
@@ -5,13 +5,14 @@ const NODES_SET = 'colyseus:nodes';
 const DISCOVERY_CHANNEL = 'colyseus:nodes:discovery';
 
 export interface Node {
-    port: number;
-    processId: string;
+  port: number;
+  processId: string;
 }
 
 async function getNodeAddress(node: Node) {
-  const ipv4 = await ip.v4();
-  return `${node.processId}/${ipv4}:${node.port}`;
+  const host = process.env.SELF_HOSTNAME || await ip.v4();
+  const port = process.end.SELF_PORT || node.port;
+  return `${node.processId}/${host}:${port}`;
 }
 
 export async function registerNode(presence: Presence, node: Node) {


LukeWood:
Hey endel - anything I need to do to get this submitted?

src/discovery/index.ts
```diff
@@ -5,13 +5,31 @@ const NODES_SET = 'colyseus:nodes';
...
+  // Default to dns resolution
+  return await ip.v4();
+}
+
+function getPort(node: Node) {
+  if (process.env.GET_HOSTNAME_FROM === 'env') {
+    return process.env.SELF_PORT || node.port;
+  }
+
+  return node.port;
```
endel, line 27: Hi @LukeWood, couldn't we simply use `process.env.SELF_PORT || node.port`, and `process.env.SELF_HOSTNAME  || await ip.v4()`?

I think that'd be easier to understand.

src/discovery/index.ts
```diff
@@ -5,13 +5,31 @@ const NODES_SET = 'colyseus:nodes';
...
 }
 
 async function getNodeAddress(node: Node) {
-  const ipv4 = await ip.v4();
-  return `${node.processId}/${ipv4}:${node.port}`;
+  const host = await getNodeHost();
+  return `${node.processId}/${host}:${getPort(node)}`;
```
endel, line 32: We could put both of them here, without a separate function for each of them

LukeWood: done

LukeWood:
updated to not have the new functions - no need to have that top level environment variable good point there.

Maybe if we have several mehods to get hostnames but seems overkill for now

endel:
Looking great, thanks for this @LukeWood 🎉 👏

PR #346, opened 2020-06-05
The first patch should be the full state logically.

chsqn:


chsqn:
Currently, the full state (encodeAllFiltered/encodeAll) and the first patch (encodeFiltered/encode) are same, so clients will receive same data twice.

PR #351, opened 2020-06-19
Fix sending default state.

chsqn:


endel:
Hi @chsqn, is this change really necessary for you? I can't recall exactly what went wrong when I've tried this, but it caused problems in a project I've tried out.

chsqn:
No it is not necessary, the code can be moved to onAuth.

chsqn:
When does the colyseus-defold 0.14 release? I can take a try. I guess there is no need to set defaults in schema constructor.

endel:
Hi @chsqn, I forgot to create the 0.14.0 tag on GitHub for the Defold client, it is available here now: https://github.com/colyseus/colyseus-defold/releases/tag/0.14.0

Please let me know if you face any issues!

GreenArrow18:
Hi @endel, is the new 0.14 Defold client compatible with the new websocket extension?

endel:
Hi @GreenArrow18, not yet, afaik the new [websocket-extension from Defold](https://github.com/defold/extension-websocket) is still somewhat in beta (as it still relies on the beta version of Defold)

With very few changes it is possible to migrate to the new websocket-extension though (code on this branch: https://github.com/colyseus/colyseus-defold/tree/websocket-extension, the `master` branch needs to be merged into the `websocket-extension` branch, as the later is still on version 0.13)

I think we need more testing with the new websocket extension, specially on iOS/Android, as it doesn't seem to have been thoroughly tested on many targets

chsqn:
Defaults in root schema constructor will be overwritten (not send to client) by events that occur before the first state patch send out.
I think this is not neither a bug nor a feature.
The PR would be closed.

Issue #396, opened 2021-02-05
Colyseus crash at Decoder.parse


Zglorb:
## Steps to Reproduce (for bugs)
1. Send a message to the room from client (via Websocket)
2. Intercept the websocket request exemple:

> ¦answer£asw°dsqdsqdqscxwdqqs¥qstNb

3. modify and resend the request to server (exemple of request sent)

> ¦answer

4.Server crashed:

>  Error: Could not parse
  at Decoder.parse (/app/server/colyseus/node_modules/notepack.io/lib/decode.js:255:9)
  at Decoder.map (/app/server/colyseus/node_modules/notepack.io/lib/decode.js:24:16)
  at Decoder.parse (/app/server/colyseus/node_modules/notepack.io/lib/decode.js:74:19)
  at Object.decode (/app/server/colyseus/node_modules/notepack.io/lib/decode.js:260:25)
  at MyRoom._onMessage (/app/server/colyseus/node_modules/colyseus/lib/Room.js:440:41)
  at WebSocket.emit (events.js:314:20)
  at WebSocket.EventEmitter.emit (domain.js:483:12)
  at Receiver.receiverOnMessage (/app/server/colyseus/node_modules/ws/lib/websocket.js:800:20)
  at Receiver.emit (events.js:314:20)
  at Receiver.EventEmitter.emit (domain.js:483:12)

## Context
I have an application that need to go live next week and i searched for ways to crash my server and with burp and intercepting WS i can do it almost all the time.
That's my first time reporting an error sorry if i did something wrong

## Your Environment
* Colyseus Version: ^0.13.0

PR #401, opened 2021-03-11
Websockets vuln fix

Zglorb:
Here is a fix for a vulnerability i mentioned a while ago #396


endel:
Nice one @Zglorb, thanks for reporting and for the PR! 🎉🎉

PR #402, opened 2021-03-11
WIP: SQL support driver based on Sequelize ORM

vitalyrotari:
Good evening, everyone!
This is early preview for SQL driver, it was tested with MySQL 8.0. 

Any proposals and improvements are welcome ;) 

**Packages required:** 

```
sequelize: ^6.5.0
ulid: ^2.3.0
```

**MySQL table dump:**

```sql
CREATE TABLE `room_caches` (
  `id` bigint unsigned NOT NULL AUTO_INCREMENT,
  `uuid` char(26) CHARACTER SET utf8mb4 COLLATE utf8mb4_bin NOT NULL,
  `name` varchar(255) COLLATE utf8mb4_general_ci NOT NULL,
  `process_id` varchar(255) COLLATE utf8mb4_general_ci NOT NULL,
  `room_id` varchar(255) COLLATE utf8mb4_general_ci NOT NULL,
  `clients` int unsigned DEFAULT '0',
  `max_clients` int DEFAULT NULL,
  `metadata` json DEFAULT NULL,
  `locked` tinyint(1) DEFAULT '0',
  `private` tinyint(1) DEFAULT '0',
  `unlisted` tinyint(1) DEFAULT '0' COMMENT 'Used for default LobbyRoom (prevent from showing up on room listing)',
  `created_at` datetime NOT NULL,
  `updated_at` datetime NOT NULL,
  PRIMARY KEY (`id`),
  UNIQUE KEY `uuid` (`uuid`),
  KEY `room_caches_name_locked` (`name`,`locked`),
  KEY `room_caches_locked` (`locked`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci;
```

### DIFF
--- a/src/matchmaker/drivers/SequelizeDriver/Query.ts
+++ b/src/matchmaker/drivers/SequelizeDriver/Query.ts
@@ -1,0 +1,43 @@
+import { SortOptions } from '../../RegisteredHandler';
+import { QueryHelpers } from '../Driver';
+import { Model, Filterable, Order, OrderItem } from 'sequelize';
+import { RoomCache } from './RoomCache';
+
+const DESC_RE = /^$(-1|desc|descending)/i
+
+export class Query<T> implements QueryHelpers<T> {
+  private conditions: Filterable<Model['_attributes']>['where'];
+  private order: Order;
+
+  constructor(conditions: any) {
+    this.conditions = conditions;
+  }
+
+  public sort(options: SortOptions) {
+    const fields = Object.entries(options);
+
+    if (fields) {
+      const order: Order = [];
+
+      for (let [field, direction] of Object.entries(options)) {
+        if (DESC_RE.test(String(direction))) {
+          order.push([field, 'DESC']);
+        } else {
+          order.push([field, 'ASC']);
+        }
+      }
+
+      this.order = order;
+    }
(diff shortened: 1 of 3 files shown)


endel:
Interesting, thanks a lot for your PR @vitalyrotari, I'm starting to think it would make sense to extract each driver into its own node module (LocalDriver having no dependencies, MongooseDriver depending on `mongoose`, yours depending on `sequelize`, etc).

I'm going to review this soon, and hopefully come up with a way to extract this from the core

vitalyrotari:
@endel Yep, the best way is to export drivers to separate packages. In our project, we were forced to translate everything into Mongo only for the driver, and so I found a little time and threw a driver for MySQL/PostgresSQL/etc.

oyed:
If possible, could you use Sequelize migrations for this? Otherwise I don't think an OEM is needed for a single table, could get away with using Knex or something

vitalyrotari:
@oyed The reason why Sequelize is very simple, from the start it was made based on the mongoose, and MatchMaker use property assignment to document and then save. The target was to make the driver without changing the core. I don't see any problems to use Sequelize.

Issue #340, opened 2020-05-14
Overwrite generateId


confuser:
Currently building a game where players join rooms with their friends only. I've currently built this to use the roomId as the key they share in order to join. 

However the current id generation is case sensitive and adds additional characters such as _ and - which in circumstances where they're not able to copy/paste easily (people playing in the same room but on different devices etc) is somewhat problematic from a UX perspective. 

Would it be possible to overwrite the generateId function for room creation? https://github.com/ai/nanoid#custom-alphabet-or-size looks ideal, but would be more flexible if this could support non nanoid alternatives as well.

MateusMendesSantana:
Hi. You can override the room id at the time of creation ;D
```
async onCreate(options: any) {
  this.roomId = nanoid()
}
```

PR #404, opened 2021-03-13
Allow specifying roomId during room creation

tobspr:
The matchmaker currently generates the `roomId` for every created room. In combination with a separate matchmaker, it can be required to be able to specify the `roomId` in advance.

This PR adds a new optional parameter to the `createRoom` method, which allows to specify a room id. If the parameter is not set, everything behaves as before.

Once there are more parameters, it might make sense to have a separate options object, e.g.:
```js
export async function createRoom(roomName: string, clientOptions: ClientOptions, { roomId } : {roomId?: string}): Promise<RoomListingData> {
}
```

### DIFF
--- a/src/MatchMaker.ts
+++ b/src/MatchMaker.ts
@@ -212,7 +212,7 @@ export function hasHandler(name: string) {
 /**
  * Create a room
  */
-export async function createRoom(roomName: string, clientOptions: ClientOptions): Promise<RoomListingData> {
+export async function createRoom(roomName: string, clientOptions: ClientOptions, roomId?: string): Promise<RoomListingData> {
   const roomsSpawnedByProcessId = await presence.hgetall(getRoomCountKey());
 
   const processIdWithFewerRooms = (
@@ -225,7 +225,7 @@ export async function createRoom(roomName: string, clientOptions: ClientOptions)
 
   if (processIdWithFewerRooms === processId) {
     // create the room on this process!
-    return await handleCreateRoom(roomName, clientOptions);
+    return await handleCreateRoom(roomName, clientOptions, roomId);
 
   } else {
     // ask other process to create the room!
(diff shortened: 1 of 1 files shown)


darrenswhite:
@tobspr I've been looking to do something similar. Although I would like to specify roomId as a function (server-side). This would allow overriding the default id generation.

darrenswhite:
> Hi @darrenswhite, this PR introduces an internal API call change, you're probably looking for overriding the `this.roomId` upon room creation, which is possible to do [as described here](https://github.com/colyseus/colyseus/issues/340#issuecomment-629720696) (just need to make sure the id you're specifying is unique!)

Hey @endel, thanks for the reply!

How would I go about ensuring a unique id? Is there a built-in mechanism for doing this? Or is it as simple as checking that a room does not exist with the given id (if this is the case how can I query a room by id?).

banool:
@darrenswhite See some code showing how I do it:
```
const LETTERS = "ABCDEFGHIJKLMNOPQRSTUVWXYZ";

export class MyRoom extends Room<MyRoomState> {
    LOBBY_CHANNEL = "$mylobby"

    generateRoomIdSingle(): string {
        var result = '';
        for (var i = 0; i < 4; i++) {
            result += LETTERS.charAt(Math.floor(Math.random() * LETTERS.length));
        }
        return result;
    }

    async generateRoomId(): Promise<string> {
        // Technically this doesn't deal with the case where all IDs are taken,
        // but we'll never have 15 quadrillion games at once.
        const currentIds = await this.presence.smembers(this.LOBBY_CHANNEL);
        var id;
        while (true) {
            id = this.generateRoomIdSingle();
            if (!currentIds.includes(id)) {
                return id;
            }
        }
    }

    async onCreate(options:any) {
        // Generate room ID, set it, and add it to the channel
        // via the Presence API (so the ode can't be re-used).
        this.roomId = await this.generateRoomId();
        this.presence.sadd(this.LOBBY_CHANNEL, this.roomId);
    }
```

This code shows you how to generate a new code (4 upper case letters in this case) and then register the code with the presence API to ensure the code doesn't get used again (probably, disregarding a very unlikely race).

@endel Would you accept a PR to add a page to the How-To section of the docs like "How to change the room ID generation logic" with this code (cleaned up) and an explanation?

endel:
Hi @banool, if you can, please do! A PR is very welcome for this!

banool:
PR here: https://github.com/colyseus/docs/pull/60.

endel:
Hi @tobspr, thank you for taking your time to create this PR!

Your additions are valuable but I think the API could be designed differently - we currently lack a "low level" matchmaking API to enable you to freely write your own matchmaking (as you did yourself!) I think we can iterate on this idea and re-organize our server-side matchmaking API somehow.

Most of the server-side matchmaking methods are directly tied to the client-side calls. IMO adding `roomId` as the third parameter on `createRoom` can add a bit of confusion. I'd be happy if both `roomId` (for `createRoom()`) and `sessionId` (for `reserveSeatFor()`) were accepted as `options` on the second parameter, but that would give too much power for the client-side - as these `options` are provided by the client. (Maybe the direct connection between the client-side calls to the equivalent method on the server-side is the real problem here!) 

I'm going to close this, but feel free to open a new issue if you'd like to contribute with ideas for a future API! We're going to think about this as well!

PR #411, opened 2021-05-26
Make `room.broadcastPatch()` public.

rnd256:
So that `Command`s (https://github.com/colyseus/command) can call it
from inside `execute()`. This is useful for turn based games where you
only ever want to broadcast a patch after a client has sent a command to
the server.


endel:
Thanks!

Issue #330, opened 2020-04-28
afterNextPatch feature: send and broadcast need to be in a queue


dengzhaofun:
Because the official does not fully implement the afterNextPath feature, I wrote my own implementation, hoping that the official version of send and broadcast are in a queue to ensure the orderly distribution of messages.

```typescript

export default class BaseRoom extends Room {
  private _afterNextPatch: Array<{type: string, args: [any, any, IBroadcastOptions]}> = [];

  constructor() {
    super()
    this['broadcastAfterPatch'] =   () => {
      const length = this._afterNextPatch.length;
      if (length > 0) {
        for (let i = 0; i < length; i++) {
          const value = this._afterNextPatch[i]
          if (value.type === 'send' && _.indexOf(this.clients, value.args[0]) !== -1) {
            this.send.apply(this, value.args)
          } else if (value.type === 'broadcast') {
            this.broadcast.apply(this, value.args);
          }
        }
        // new messages may have been added in the meantime,
        // let's splice the ones that have been processed
        this._afterNextPatch.splice(0, length);
      }
    }
  }
   delaySend(client: Client, type: string | number, message: any) {
    this._afterNextPatch.push({ type: 'send', args: [client, type, message]})
  }

  delayBroadcast(type: string | number, message: any, options: IBroadcastOptions = {}) {
    if (options.afterNextPatch) {
      delete options.afterNextPatch;
    }
    this._afterNextPatch.push({ type: 'broadcast', args: [type, message, options]})
  }
}
```

endel:
This is great @dengzhaofun, thanks a lot for taking the time to implement this. I'm going to adapt it to release this on `0.13.2` soon!

PR #416, opened 2021-06-15
After next patch

DamithaPerera-Allion:


Wenish:
https://github.com/colyseus/colyseus/issues/330

PR #438, opened 2021-08-27
fixes #431 Package publisher workflow

### LINKED ISSUE #431: GitHub Workflow: publishing changed packages to NPM registry
Now that we're using a monorepo & Lerna (https://github.com/colyseus/colyseus/pull/419) to manage internal Colyseus' packages, we need to establish a way to auto-publish updated packages on NPM.

As we are using [independent mode](https://github.com/lerna/lerna#independent-mode), each package can evolve on its own versioning numbers. Every `0.14.x` version should be compatible with each other. Once we introduce a breaking change - the core and all its internal packages are going to be bumped to the next version (e.g. `0.15.x`)

---

Lerna built-in versioning and publishing tools seems to be centered around a "fixed" version number for every package, not really friendly with the "independent" mode.

- Because of that, I'm currently having to manually run `npm publish` for each package that has been changed
  - _(a GitHub Workflow could handle this)_
- I manually update the `"version"` number of `package.json` for that package, too. 
  - _(This step is fine, it gives us control)_

As you can see, the `colyseus` bundle package uses `^` for specifying the version number of its dependencies. So if an internal package gets updated, (e.g. `"@colyseus/ws-transport"` is bumped to `0.14.99`) - it is optional for consumers if they want to upgrade it or not.
https://github.com/colyseus/colyseus/blob/a404b0fe9113392ed4c1cae7918ccd79aedc0350/bundles/colyseus/package.json#L34-L37

---

This thread has some useful information on approaches to take for "independent" versioning and CI: https://github.com/lerna/lerna/issues/1286

lpsandaruwan:
This workflow will check for version changes comparing the version with published artifacts in NPM package registry and will publish only the version changed packages.

A NPM token(for publishing package) need to be added to the Github repository secrets(named as `NPM_TOKEN`)  to use this CI workflow.
To generate NPM token, on [https://www.npmjs.com/](https://www.npmjs.com/)
1. In the upper right corner of the page, click your profile picture, then click Access Tokens.
2. Screenshot of the account menu with the tokens link selected
3. Click Create New Token.
4. Select the type of access token: Automation
5. Click Generate Token.
6. Copy the token from the top of page.
As explained in [https://docs.npmjs.com/creating-and-viewing-access-tokens](https://docs.npmjs.com/creating-and-viewing-access-tokens)

Then add it to the Github, as explained in [https://docs.github.com/en/actions/reference/encrypted-secrets](https://docs.github.com/en/actions/reference/encrypted-secrets)

1. On GitHub, navigate to the main page of the repository.
2. Under your repository name, click  Settings.
3. Repository settings button
4. In the left sidebar, click Secrets.
5. Click New repository secret.
6. Type a name for your secret in the Name input box. int this case it is `NPM_TOKEN`
7. Enter the value for your secret.
8. Click Add secret.


endel:
Hi @lpsandaruwan, thanks for the PR, this approach is looking good! 

I was trying to run it locally, though, and I'm getting this error:

```
/Users/endel/Projects/colyseus/colyseus/.github/scripts/node_modules/npm/lib/view.js:61
    if (data["dist-tags"].hasOwnProperty(version)) {
                          ^

TypeError: Cannot read property 'hasOwnProperty' of undefined
    at /Users/endel/Projects/colyseus/colyseus/.github/scripts/node_modules/npm/lib/view.js:61:27
```

(Node v16.1.0, NPM 7.11.2) 

If you need more info about the error just let me know, we can try to inspect it together 

---

A few suggestions that I have:
- Use `NPM_AUTH_TOKEN` env var instead of `NPM_TOKEN` (we already have an `NPM_AUTH_TOKEN` set for publishing 🙂)
- Skip "private" packages during the version comparison loop (some packages are either internal and/or not ready and flagged as "private", [such as the example package](https://github.com/colyseus/colyseus/blob/bdf43169e509e0fd2305cd4a3b3669e7bf733c76/packages/example/package.json#L3))

lpsandaruwan:
Hi @endel,
Thank you very much! I will investigate the above,  also will look for applying  your suggestions and will get back to you. :+1

lpsandaruwan:
Hi @endel,

The above error you got due to an old package `npmview`, I removed that dependency from workflow. Also added functionality to skip private packages and changed the NPM token variable name. Please review and let me know if it requires more changes.

Issue #439, opened 2021-08-27
Allow to customize Access-Control-Allow-Headers passed to matchmake endpoint


JCMais:
The value is currently static here: https://github.com/colyseus/colyseus/blob/bdf43169e509e0fd2305cd4a3b3669e7bf733c76/packages/core/src/Server.ts#L236-L247

If you use an instrumentation library (such as newrelic) it may add new headers , right now this is broken

## Steps to Reproduce (for bugs)
1. Enable any instrumentation library that add new headers to allow for CORS tracing 
2. Matchmake does not work anymore as we requested more headers than what was returned

See screenshot:

![image](https://user-images.githubusercontent.com/1041647/131159981-d3766f39-78cc-45ec-b202-931418d7ca05.png)

Notice the requested headers (`Access-Control-Request-Headers`) and the ones returned in the response (`Access-Control-Allow-Headers`)

## Context

Trying to enable the NewRelic browser monitoring agent

## Your Environment

* Colyseus Version: v0.14.20
* Node.js Version: v14
* Operating System and version: Ubuntu

endel:
Hi @JCMais, this has been taken care of on the upcoming 0.15 version 🤞  https://github.com/colyseus/colyseus/blob/3d6624239fa89e856aea2d1b82ac49d8647aedbb/packages/transport/uwebsockets-transport/src/uWebSocketsTransport.ts#L179

PR #440, opened 2021-08-30
#437 Return all rooms when roomName is undefined

### LINKED ISSUE #437: getAvailableRooms (roomName?: string)
In the doc, If the roomName parameter is ommitted, all rooms are going to be queried.
But in fact, none of rooms returned if the parameter is null.

lpsandaruwan:
When `roomName ` is not provided, `roomName` has been taken as `matchmake` from the request URI. This PR is fixes #437

### DIFF
--- a/bundles/colyseus/test/IntegrationTest.ts
+++ b/bundles/colyseus/test/IntegrationTest.ts
@@ -886,6 +886,29 @@ describe("Integration", () => {
 
           })
 
+          describe("Matchmaker queries", () => {
+            const createDummyRooms = async () => {
+              matchMaker.defineRoomType('allroomstest', class _ extends Room {});
+              matchMaker.defineRoomType('allroomstest2', class _ extends Room {});
+              await matchMaker.create("allroomstest");
+              await matchMaker.create("allroomstest2");
+            }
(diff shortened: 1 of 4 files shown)


endel:
Interesting, thanks for the PR @lpsandaruwan! Looking good

- Could you write a small integration test for this? Integration tests are located at `bundles/colyseus/test/IntegrationTest.ts`
- Can you dup this change here as well? 
https://github.com/colyseus/colyseus/blob/master/packages/transport/uwebsockets-transport/src/uWebSocketsTransport.ts#L241-L249

This piece of code is duplicated for `uWebSocketsTransport` - eventually, it would be nice to unify both implementations somehow, as there are other small issues that could benefit from a single implementation (e.g. https://github.com/colyseus/colyseus/issues/439) - the problem with uWebSockets.js is that it does not use the built-in Node.js Server APIs, making it a bit difficult to have a single implementation

Feel free to reach out if you need any information! Cheers

lpsandaruwan:
Hi @endel,

Fixed the code as you mentioned and also added the integration test. Please review and let me know if there requires any change.
Thank you! :+1:

PR #447, opened 2021-09-07
feat: export registerSerializer

### LINKED ISSUE #446: Register serializer for testing library
## Context

At my company we use a different serializer than `SchemaSerializer` because we have integration with a mobile app, everything works fine but we currently can't utilize the testing library because `registerSerializer` function isn't exported.

It will be really great if we can register a new serializer for the client used by the testing library

## Your Environment

* Colyseus Testing library Version: 0.14.22 
* Node.js Version: 14

YoussefEgla:
Resolve #446

### DIFF
--- a/packages/testing/src/TestServer.ts
+++ b/packages/testing/src/TestServer.ts
@@ -1,5 +1,5 @@
 import { Server, Room, matchMaker } from "@colyseus/core";
-import { Client } from "colyseus.js";
+import { Client, registerSerializer } from "colyseus.js";
 import * as httpie from "httpie";
 
 export class ColyseusTestServer {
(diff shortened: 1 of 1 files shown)


endel:
Thanks for the PR @YoussefEgla,

I'm wondering if you couldn't import and use `registerSerializer` inside your test files, though? The registered serializer are global and don't need a specific scope in order to work:

```typescript
import { registerSerializer } from "colyseus.js";

// register your serializer
registerSerializer(...);

describe(() => {
  it("your test...", () => {
    // ...
    // assert.ok()
    
  });
})
```

If that works for you I'd prefer not to expose it through the testing library at this time.

YoussefEgla:
Unfortunatley, that didnt work and error out. Also I forgot to mention this is for writing tests for the backend

PR #452, opened 2021-09-16
use "ioredis" instead of "redis" for Cluster support

endel:
This PR replaces the `redis` package with `ioredis` for both `@colyseus/redis-presence` and `@colyseus/redis-driver`.

### DIFF
--- a/packages/drivers/redis-driver/src/index.ts
+++ b/packages/drivers/redis-driver/src/index.ts
@@ -1,4 +1,4 @@
-import redis, { RedisClient, ClientOpts } from 'redis';
+import Redis from 'ioredis';
 import { promisify } from 'util';
 
 import {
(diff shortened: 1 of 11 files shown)


packages/drivers/redis-driver/src/index.ts
```diff
@@ -12,12 +12,10 @@ import { Query } from './Query';
...
 export class RedisDriver implements MatchMakerDriver {
-  private readonly _client: RedisClient;
-  private readonly hgetall: (key: string) => Promise<{ [key: string]: string }>;
+  private readonly _client: Redis.Redis;
 
-  constructor(options?: ClientOpts, key: string = 'roomcaches') {
-    this._client = redis.createClient(options);
-    this.hgetall = promisify(this._client.hgetall).bind(this._client);
+  constructor(options?: Redis.RedisOptions, key: string = 'roomcaches') {
+    this._client = new Redis(options);
```
epiphone, line 18: Don't we need to use the [`Cluster`](https://github.com/luin/ioredis#cluster) interface instead to handle clusters? Connecting to a Redis cluster configuration endpoint with `ioredis.Redis` seems to throw `ReplyError: MOVED`.

epiphone: Here's a PR for cluster support in the proxy: https://github.com/colyseus/proxy/pull/25

Issue #430, opened 2021-08-19
Proposal for @colyseus/loadtest


endel:
The `colyseus-loadtest` tool currently requires the input file to follow its own "DSL", implementing the methods `onJoin`, `onLeave`, `onStateChange`, etc.

**Example** ([See full example](https://github.com/colyseus/colyseus/blob/a404b0fe9113392ed4c1cae7918ccd79aedc0350/packages/loadtest/example/bot.ts#L1-L25))

```typescript
// ...
export function requestJoinOptions (this: Client, i: number) {
    return { requestNumber: i };
}

export function onJoin(this: Room) {
    console.log(this.sessionId, "joined.");

    this.onMessage("*", (type, message) => {
        console.log("onMessage:", type, message);
    });
}

export function onLeave(this: Room) {
    console.log(this.sessionId, "left.");
}
// ...
```

## Proposed behaviour

No "DSL": It would be more intuitive (and less learning curve required) to allow the user to provide a plain script file that simply consumes the regular `colyseus.js` package. By overwriting/monkey-patching `colyseus.js`'s internal methods we can keep tracking bandwidth and memory usage by the client while making it easier to write "bot" scripts.

Multi-threading: as Node.js is single-threaded, the more concurrent connections spawned through the loadtest tool, less responsive the Node.js process will become - making some client connections drop because of that. Depending on the CPU, 300 connections are enough to start losing client connections.

**Example / source-code:**

```typescript
import { Client } from "colyseus.js";

export async function main(options: any) {
  const client = new Client(options.endpoint);
  const room = await client.joinOrCreate(options.room);
  room.onMessage("msg-type", (payload) => {
    // ...
  });
  room.onLeave((code) => {
    // ...
  });
}
```

endel:
Just leaving this as a reference... On `@colyseus/testing` the monkey-patching has been done this way, we can do similarly for `colyseus.js` during the loadtest: 
https://github.com/colyseus/colyseus/blob/8dd5b6407122cdcab362f5f534bad68eef1dda82/packages/testing/src/Room.ext.ts#L29-L35

endel:
Landed on 0.15 https://docs.colyseus.io/tools/loadtest/#example

Multi-threading is still not supported though.

PR #453, opened 2021-09-27
loadtest improvements 

lpsandaruwan:
Solves #430 
Usage: Commands are same as before.
Options object in bot script consisting,
```
    endpoint: argv.endpoint || `ws://localhost:2567`,
    roomName: argv.room,
    roomId: argv.roomId,
    numClients: argv.numClients || 1,
    scriptFile: argv._[0] && path.resolve(argv._[0]),
    delay: argv.delay || 0,
    logLevel: argv.logLevel?.toLowerCase() || "all", // TODO: not being used atm
    reestablishAllDelay: argv.reestablishAllDelay || 0,
    retryFailed: argv.retryFailed || 0,
    output: path.resolve(argv.output || "loadtest.log"),
    requestJoinOptions: {}
```


endel:
Looking great @lpsandaruwan 👏 

---

I've left some comments in the PR, to allow the `Options` type we may require a custom build step, since the `"module"` and `"main"` on `package.json` can point to an empty file (the entry point was never really used, as we only use the binary)

lpsandaruwan:
Thanks @endel, I will work on those. :-)

PR #454, opened 2021-10-04
Abstract logging support

### LINKED ISSUE #442: Custom logger support
Original report by @MCArth:

>  what might be an interesting addition to the engine would be the ability to set logging levels of colyseus (and then there be an 'error only' level which excludes this/seat reservation expired that could be called warnings or something)
> but that's a lot of work and tbh I should just set up my own logging system haha

---

The disadvantage of using `console.log` directly inside the Colyseus' codebase (which we currently do) is that all internal logs are forcibly going to go to stdout, making it not possible to easily forward logs and errors to external services.

We can do exactly as the web framework [fastify](https://github.com/fastify/fastify): use a small [abstract-logging](https://www.npmjs.com/package/abstract-logging) library for logs, and default it to `console` to keep current functionality.

There are quite a few logging frameworks available for Node.js, such as [`winston`](https://www.npmjs.com/package/winston), [`bunyan`](https://www.npmjs.com/package/bunyan), [`pino`](https://www.npmjs.com/package/pino), etc. By using an abstract logger, end-users could provide whichever logging library they prefer:

**Example providing a `pino` logger:**

```typescript
import pino from "pino";
const transport = pino.transport({
  target: 'some-transport',
  options: { some: 'options for', the: 'transport' }
})
const logger = pino(transport);

const server = new Server({
  // provide the custom logger. 
  // default: console
  logger: logger,
});
```

lpsandaruwan:
Resolves #442 
* Can be configured via server options,
i.e.
```
import Arena from "/home/lahirup/Desktop/Colyseus/colyseus/packages/arena/build";
import { monitor } from "/home/lahirup/Desktop/Colyseus/colyseus/packages/monitor/src";

import * as winston from "winston";

const LOG: winston.Logger = winston.createLogger({
    format: winston.format.combine(
        winston.format.timestamp(),
        winston.format.json(),
    ),
    level: 'info',
    transports: [
        new winston.transports.File({ filename: 'error.log', level: 'error' }),
        new winston.transports.File({ filename: 'all.log' }),
    ],
});

import { TicTacToeRoom } from "./rooms/TicTacToeRoom";

export default Arena({
    options: {
        logger: LOG
    },
    getId: () => "Your Colyseus App",
    initializeGameServer: (gameServer) => {
        gameServer.define('tic_tac_toe_game_room', TicTacToeRoom);
    },

    initializeExpress: (app) => {
        app.get("/", (req, res) => {
            res.send("Server ready!");
        });

        app.use("/colyseus", monitor());
    },

    beforeListen: () => {
    }
});
```

* Usage
```
import { Room, Client, Logger as LOG } from "/home/lahirup/Desktop/Colyseus/colyseus/packages/core/build";
import { TicTacToeRoomState } from "./schema/TicTacToeRoomState";

export class TicTacToeRoom extends Room<TicTacToeRoomState> {

  maxClients = 2;

  onCreate (options: any) {
    ...
  }

  onJoin(client: Client, options: any) {
    LOG.info(client.sessionId, "joined!");
  }

  onLeave (client: Client, consented: boolean) {
    LOG.info(client.sessionId, "left!");
  }

  onDispose() {
    LOG.info("room", this.roomId, "disposing...");
  }
}
```

### DIFF
--- a/packages/arena/src/index.ts
+++ b/packages/arena/src/index.ts
@@ -4,10 +4,7 @@ import path from "path";
 import cors from "cors";
 import express from "express";
 import dotenv from "dotenv";
-import {Logger, Server, ServerOptions, Transport} from '@colyseus/core';
-
-// system logger
-const log = Logger.getLogger();
+import {Logger as LOG, Server, ServerOptions, Transport} from '@colyseus/core';
 
 // try to import uWebSockets-express compatibility layer.
 let uWebSocketsExpressCompatibility: any;
@@ -31,9 +28,9 @@ if (process.env.NODE_ARENA !== "true") {
 
     if (envPath) {
         dotenv.config({ path: envPath });
-        log.info(`✅ ${envFilename} loaded.`);
+        LOG.info(`✅ ${envFilename} loaded.`);
     } else {
-        log.info(`⚠️  ${envFilename} not found.`);
+        LOG.info(`⚠️  ${envFilename} not found.`);
     }
 }
 
@@ -95,10 +92,10 @@ export async function listen(
     if (options.displayLogs) {
         const appId = options.getId?.() || "[ Colyseus ]";
         if (appId) {
-            log.info(`🏟  ${appId}`);
+            LOG.info(`🏟  ${appId}`);
         }
 
-        log.info(`⚔️  Listening on ws://localhost:${port}`);
+        LOG.info(`⚔️  Listening on ws://localhost:${port}`);
     }
     return gameServer;
 }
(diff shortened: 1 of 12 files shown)


endel:
Looking great @lpsandaruwan, although I'm going to nitpick a bit 😄 let's go:

Static classes as you've done are an OOP anti-pattern. I suggest not using any static methods or properties for `Logger`, and hold the logger instance at the module level instead. 

```typescript
export class Logger {
  info() {}
  log() {}
  // etc ...
}

// Use "console" as our default logger
export let logger: Logger = console;

export function setLogger(instance: Logger) {
  logger = instance;
}
```

This way we can import the `logger` instance directly from anywhere inside the Colyseus core.

```typescript
// Use this
import { logger } from "./Logger";
logger.warn(...)

// instead of this
import { Logger as LOG } from "./Logger";
LOG.warn(...)
```

We can remove the `Logger.getLogger()` method by going this way.

Let me know if you have any thoughts or questions. Cheers!

lpsandaruwan:
Hi @endel,
I changed the static properties and functions as you explained, tested the code and updated the PR. Thanks for the guidance. :+1:  :-)

PR #473, opened 2021-11-09
Reconnection using a token

### LINKED ISSUE #354: Reconnection: use a private "reconnection token" instead of "sessionId"
There is a security vulnerability that could allow any client that has the `sessionId` and `roomId` to "reconnect" and act on other client's behalf in case he got disconnected AND `allowReconnection()` is being used with his `sessionId`.

lpsandaruwan:
Resolves #354


endel:
Thanks @lpsandaruwan but we're moving with a slightly simpler approach here  https://github.com/colyseus/colyseus/pull/484

PR #479, opened 2021-11-15
#354: reconnection with private reconnection token

mmosiura:


endel:
Thanks @mmosiura, took some inspiration from your implementation here https://github.com/colyseus/colyseus/pull/484, soon available on version `0.15`

PR #504, opened 2022-02-15
Redis Driver namespacing

epiphone:
`RedisDriver` already accepts a `key: string` argument but the argument is not used anywhere: instead the class uses a hardcoded `roomcaches` value.

This PR changes updates `RedisDriver` so that if `key` is defined, use the argument value instead of the hardcoded `roomcaches` value. This is a non-breaking change since `key`'s default value matches the previously hardcoded value.

Namespacing like this is useful since it allows sharing a single Redis instance between multiple Colyseus instances.


Jakedavies:
This would be a breaking change for anyone setting this key I assume as it would reset the cache.

epiphone:
@endel should I rebase this against https://github.com/colyseus/colyseus/pull/452? The test mocks need updating to handle `ioredis`.

endel:
Hi @epiphone, thank you for the PR (and the other one on the proxy!) gonna review and merge both soon!

https://github.com/colyseus/colyseus/pull/452 has been merged on the [`0.15` branch](https://github.com/colyseus/colyseus/tree/0.15), if you can rebase against https://github.com/colyseus/colyseus/pull/452 would make it easier for merging this into `0.15` too!

epiphone:
Thanks @endel! Now pointing to `ioredis`.

PR #512, opened 2022-03-23
Dev Mode

lpsandaruwan:
Enables dev mode when setting `DEV_MODE=true` env variable.
Related to: https://github.com/colyseus/schema/issues/115

### DIFF
--- a/packages/core/src/MatchMaker.ts
+++ b/packages/core/src/MatchMaker.ts
@@ -37,7 +37,7 @@ export interface SeatReservation {
   room: RoomListingData;
 }
 
-const handlers: {[id: string]: RegisteredHandler} = {};
+export const handlers: {[id: string]: RegisteredHandler} = {};
 const rooms: {[roomId: string]: Room} = {};
 
 const devMode: boolean = Boolean(process.env.DEV_MODE);
@@ -446,7 +446,7 @@ async function cleanupStaleRooms(roomName: string) {
   }));
 }
 
-async function createRoomReferences(room: Room, init: boolean = false): Promise<boolean> {
+export async function createRoomReferences(room: Room, init: boolean = false): Promise<boolean> {
   rooms[room.roomId] = room;
 
   if (init) {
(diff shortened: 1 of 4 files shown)


packages/core/src/Utils.ts
```diff
@@ -259,3 +273,118 @@ export declare interface DummyServer {
 }
 
 export class DummyServer extends EventEmitter {}
+
+export async function reloadFromCache() {
```
endel, line 277: This method is mostly duplicated from ` handleCreateRoom()`. I suggest moving this method to `MatchMaker.ts` and call `handleCreateRoom` instead, possibly passing a third argument to specify the room is being "re-loaded".

packages/core/src/Utils.ts
```diff
@@ -259,3 +273,118 @@ export declare interface DummyServer {
...
+      await presence.hset(getRoomHistoryListKey(), room.roomId, JSON.stringify(roomHistory));
+    }
+  }
+}
+
+export async function getPreviousProcessId(hostname) {
+  return await presence.hget(getProcessHistoryKey(), hostname);
+}
+
+export function getRoomCountKey() {
```
endel, line 376: Let's not duplicate these methods

packages/core/src/Utils.ts
```diff
@@ -1,11 +1,25 @@
...
+import {
+  createRoomReferences,
+  disposeRoom,
+  driver, handlers,
+  lockRoom, onClientJoinRoom, onClientLeaveRoom,
+  presence,
+  processId,
+  publicAddress,
+  remoteRoomCall, unlockRoom
+} from "./MatchMaker";
```
endel, line 13: No need to import most of these, since `handleCreateRoom` and related methods should live in the `MatchMaker.ts` file. (No need to add `export` to most of them, too)

packages/core/src/MatchMaker.ts
```diff
@@ -470,27 +502,27 @@ async function awaitRoomAvailable(roomToJoin: string, callback: Function): Promi
   });
 }
 
-function onClientJoinRoom(room: Room, client: Client) {
+export function onClientJoinRoom(room: Room, client: Client) {
```
endel, line 505: No need to `export` here neither others that are used on `handleCreateRoom()`

endel (changes requested):
Hi @lpsandaruwan, I've left some comments here 🙏

endel:
Thanks @lpsandaruwan, merged this into the `0.15` branch here https://github.com/colyseus/colyseus/pull/485 🎉  

To cache `LocalPresence` on `devMode` I've took a different approach here https://github.com/colyseus/colyseus/pull/485/commits/fc75f5164a347c8e0c6387b2a7b480dfc6caeef6#diff-5062eda5482deb63ab42a1a2e69c6b07844f3c44ea1a42c563c64caaa3be683d

PR #530, opened 2022-07-25
CLYS-101 Code documentation improvments

lpsandaruwan:


packages/core/src/Room.ts
```diff
@@ -53,19 +60,59 @@ export abstract class Room<State= any, Metadata= any> {
...
 
+  /**
+   * A unique, auto-generated, 9-character-long id of the room.
+   * You may replace `this.roomId` during `onCreate()`.
+   */
   #_roomId: string;
+  /**
+   * The name of the room you provided as first argument for `gameServer.define()`.
+   */
   #_roomName: string;
```
endel, line 77: Only "userland" properties should be documented, we can move these to `roomId` and `roomName` getters. Private properties (starting with `#`) are not exposed, even when using `extends`

packages/core/src/Room.ts
```diff
@@ -87,6 +134,13 @@ export abstract class Room<State= any, Metadata= any> {
...
   private _patchInterval: NodeJS.Timer;
 
+  /**
+   * This property will change on these situations:
+   * - The maximum number of allowed clients has been reached (`maxClients`)
+   * - You manually locked, or unlocked the room using lock() or `unlock()`.
+   *
+   * @readonly
+   */
   private _locked: boolean = false;
```
endel, line 144: Move to getter instead

packages/core/src/Room.ts
```diff
@@ -336,6 +422,9 @@ export abstract class Room<State= any, Metadata= any> {
     }
   }
 
+  /**
+   * This method will check whether mutations have occurred in the state, and broadcast them to all connected clients.
```
endel, line 426: The editor already tells the user this is a method, so in cases like this we can remove `"This method will"` (or equally redundant words to describe a scope already known by the editor)

I suggest:
> "Checks whether mutations have occurred in the state, and broadcast them to all connected clients."

endel (changes requested):
Nice, thanks @lpsandaruwan! I've added a few comments 🙏

PR #536, opened 2022-08-29
Include userData in reconnect.

ryanwillis:
When client reconnects, check for and copy over `userData`.


endel:
Thank you @ryanwillis, this has been published on `@colyseus/core@0.14.32` 🥳

PR #540, opened 2022-09-30
0.15 fix async/await and code quality

vitalyrotari:
Hi, @endel 

Out project is working more than 1 year on amazing Colyseus, now we made some new feature, and I found many issues with async/await inside of core and other minor things.

I'm not sure if this PR will be merged completely, but they contains many fixes and improvements.

- Fix code style based according `tslint` config from project.
- Fix unresolved async/await 
- Unify Presence interface methods
- Unify MatchMakerDriver interface methods
- Add ClientRef interface and missing `onLeave` method
- Fix unresolved async `_reserveSeat` from `Room.allowReconnection`, so method `_reserveSeat` requires some refactoring. 

Some example of modified `allowReconnection` method

### Allow reconnection with timeout

```
  try {
    await this.allowReconnection(client, 30);
  } catch (e) {
    // ...
  }
```

### Allow reconnection with manual cancelation

```
  const controller = new AbortController();
  
  setTimeout(() => {
    controller.abort();
  }, 5000);

  try {
    await this.allowReconnection(client, 'manual', controller);
  } catch (e) {
    // ...
  }
```

### DIFF
--- a/packages/core/src/MatchMaker.ts
+++ b/packages/core/src/MatchMaker.ts
@@ -65,7 +65,7 @@ export async function setup(
   /**
    * Subscribe to remote `handleCreateRoom` calls.
    */
-  subscribeIPC(presence, processId, getProcessChannel(), (_, args) => {
+  await subscribeIPC(presence, processId, getProcessChannel(), (_, args) => {
     return handleCreateRoom.apply(undefined, args);
   });
 
@@ -165,8 +165,8 @@ export async function joinById(roomId: string, clientOptions: ClientOptions = {}
 /**
  * Perform a query for all cached rooms
  */
-export async function query(conditions: Partial<IRoomListingData> = {}) {
-  return await driver.find(conditions);
+export function query(conditions: Partial<IRoomListingData> = {}) {
+  return driver.find(conditions);
 }
 
 /**
@@ -195,7 +195,7 @@ export async function findOneRoomAvailable(roomName: string, clientOptions: Clie
       roomQuery.sort(handler.sortOptions);
     }
 
-    return await roomQuery;
+    return roomQuery;
   });
 }
 
@@ -236,7 +236,7 @@ export async function remoteRoomCall<R= any>(
   }
 }
 
-export function defineRoomType<T extends Type<Room>>(
+export async function defineRoomType<T extends Type<Room>>(
   name: string,
   klass: T,
   defaultOptions?: Parameters<NonNullable<InstanceType<T>['onCreate']>>[0],
@@ -246,17 +246,17 @@ export function defineRoomType<T extends Type<Room>>(
   handlers[name] = registeredHandler;
 
   if (!isDevMode) {
-    cleanupStaleRooms(name);
+    await cleanupStaleRooms(name);
   }
 
   return registeredHandler;
 }
 
-export function removeRoomType(name: string) {
+export async function removeRoomType(name: string) {
   delete handlers[name];
 
   if (!isDevMode) {
-    cleanupStaleRooms(name);
+    await cleanupStaleRooms(name);
   }
 }
 
@@ -306,10 +306,10 @@ export async function createRoom(roomName: string, clientOptions: ClientOptions)
   }
 
   if (isDevMode) {
-    presence.hset(getRoomRestoreListKey(), room.roomId, JSON.stringify({
-      "clientOptions": clientOptions,
-      "roomName": roomName,
-      "processId": processId
+    await presence.hset(getRoomRestoreListKey(), room.roomId, JSON.stringify({
+      clientOptions,
+      processId,
+      roomName,
     }));
   }
 
@@ -359,8 +359,7 @@ export async function handleCreateRoom(
       await room.onCreate(merge({}, clientOptions, registeredHandler.options));
 
       // increment amount of rooms this process is handling
-      presence.hincrby(getRoomCountKey(), processId, 1);
-
+      await presence.hincrby(getRoomCountKey(), processId, 1);
     } catch (e) {
       debugAndPrintError(e);
       throw new ServerError(
@@ -426,15 +425,15 @@ export async function gracefullyShutdown(): Promise<any> {
   }
 
   // remove processId from room count key
-  presence.hdel(getRoomCountKey(), processId);
+  await presence.hdel(getRoomCountKey(), processId);
 
   // unsubscribe from process id channel
-  presence.unsubscribe(getProcessChannel());
+  await presence.unsubscribe(getProcessChannel());
 
   return Promise.all(disconnectAll(
     (isDevMode)
       ? Protocol.WS_CLOSE_DEVMODE_RESTART
-      : undefined
+      : undefined,
   ));
 }
 
@@ -486,10 +485,9 @@ async function cleanupStaleRooms(roomName: string) {
     try {
       // use hardcoded short timeout for cleaning up stale rooms.
       await remoteRoomCall(room.roomId, 'roomId');
-
     } catch (e) {
       debugMatchMaking(`cleaning up stale room '${roomName}', roomId: ${room.roomId}`);
-      room.remove();
+      await room.remove();
     }
   }));
 }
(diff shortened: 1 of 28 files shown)


packages/core/src/MatchMaker.ts
```diff
@@ -65,7 +65,7 @@ export async function setup(
   /**
    * Subscribe to remote `handleCreateRoom` calls.
    */
-  subscribeIPC(presence, processId, getProcessChannel(), (_, args) => {
+  await subscribeIPC(presence, processId, getProcessChannel(), (_, args) => {
```
hunkydoryrepair, line 68: what is the point of awaiting this? Is there any reason to delay the other calls?

packages/core/src/MatchMaker.ts
```diff
@@ -165,8 +165,8 @@ export async function joinById(roomId: string, clientOptions: ClientOptions = {}
 /**
  * Perform a query for all cached rooms
  */
-export async function query(conditions: Partial<IRoomListingData> = {}) {
-  return await driver.find(conditions);
+export function query(conditions: Partial<IRoomListingData> = {}) {
```
hunkydoryrepair, line 168: this is a good improvement to make. The implementation is much better without extra layers of promises

packages/core/src/MatchMaker.ts
```diff
@@ -195,7 +195,7 @@ export async function findOneRoomAvailable(roomName: string, clientOptions: Clie
       roomQuery.sort(handler.sortOptions);
     }
 
-    return await roomQuery;
```
hunkydoryrepair, line 198: if the function is awaited by the caller, then makes more sense to just remove the `async` and not await here

packages/core/src/MatchMaker.ts
```diff
@@ -245,17 +245,17 @@ export function defineRoomType<T extends Type<Room>>(
   handlers[name] = registeredHandler;
 
   if (!isDevMode) {
-    cleanupStaleRooms(name);
+    await cleanupStaleRooms(name);
```
hunkydoryrepair, line 248: This could be breaking. Inserting asynchronous code into what is currently a syncrhonous call could break some apps, and changing to async changes the return type, so it is definitely app breaking.

packages/core/src/MatchMaker.ts
```diff
@@ -245,17 +245,17 @@ export function defineRoomType<T extends Type<Room>>(
...
   return registeredHandler;
 }
 
-export function removeRoomType(name: string) {
+export async function removeRoomType(name: string) {
   delete handlers[name];
 
   if (!isDevMode) {
-    cleanupStaleRooms(name);
+    await cleanupStaleRooms(name);
```
hunkydoryrepair, line 258: seems unnecessary. Cleaning up stale rooms should be handled asynchronously.
I think using `.catch` might be useful, but why wait?

packages/core/src/MatchMaker.ts
```diff
@@ -305,17 +305,21 @@ export async function createRoom(roomName: string, clientOptions: ClientOptions)
   }
 
   if (isDevMode) {
-    presence.hset(getRoomRestoreListKey(), room.roomId, JSON.stringify({
-      "clientOptions": clientOptions,
-      "roomName": roomName,
-      "processId": processId
+    await presence.hset(getRoomRestoreListKey(), room.roomId, JSON.stringify({
```
hunkydoryrepair, line 308: awaiting here is likely to cause timing changes that could break things. Currently, the key shows up sometime after createRoom has returned. awaiting here will cause it to show up first, which means more requests to join the room come in before createRoom has even returned.

packages/core/src/MatchMaker.ts
```diff
@@ -346,16 +350,15 @@ export async function handleCreateRoom(roomName: string, clientOptions: ClientOp
...
   });
 
   if (room.onCreate) {
     try {
       await room.onCreate(merge({}, clientOptions, registeredHandler.options));
 
       // increment amount of rooms this process is handling
-      presence.hincrby(getRoomCountKey(), processId, 1);
-
+      await presence.hincrby(getRoomCountKey(), processId, 1);
```
hunkydoryrepair, line 361: it is not important to await this unless we want to catch the error.  Of course, this might be a nice one to catch, as if REDIS is dying we could save having an unhandled exception, but using `.catch()` could be more performant.

packages/core/src/MatchMaker.ts
```diff
@@ -370,15 +373,15 @@ export async function handleCreateRoom(roomName: string, clientOptions: ClientOp
...
+  // immediately ask client to join the room
   debugMatchMaking('spawning \'%s\', roomId: %s, processId: %s', roomName, room.roomId, processId);
 
   room._events.on('lock', lockRoom.bind(this, room));
   room._events.on('unlock', unlockRoom.bind(this, room));
   room._events.on('join', onClientJoinRoom.bind(this, room));
   room._events.on('leave', onClientLeaveRoom.bind(this, room));
+
   room._events.once('dispose', disposeRoom.bind(this, roomName, room));
-  room._events.once('disconnect', () => room._events.removeAllListeners());
```
hunkydoryrepair, line 381: won't removing this leak all the room listeners? That seems pretty bad.

packages/core/src/MatchMaker.ts
```diff
@@ -421,15 +424,15 @@ export async function gracefullyShutdown(): Promise<any> {
   }
 
   // remove processId from room count key
-  presence.hdel(getRoomCountKey(), processId);
+  await presence.hdel(getRoomCountKey(), processId);
```
hunkydoryrepair, line 427: gracefullyShutdown is called as a result of unhandled exceptions. As soon as you await when processing an unhandled exception, the application will exit, thus, this will leave redis full and won't even call `disconnect`. This would be a pretty big problem when we want a chance to save data during unhandled exception.

packages/core/src/MatchMaker.ts
```diff
@@ -563,7 +565,7 @@ async function disposeRoom(roomName: string, room: Room) {
 
   // decrease amount of rooms this process is handling
   if (!isGracefullyShuttingDown) {
-    presence.hincrby(getRoomCountKey(), processId, -1);
+    await presence.hincrby(getRoomCountKey(), processId, -1);
```
hunkydoryrepair, line 568: awaiting while shutting down should be avoided

packages/core/src/MatchMaker.ts
```diff
@@ -580,10 +582,10 @@ async function disposeRoom(roomName: string, room: Room) {
   handlers[roomName].emit('dispose', room);
 
   // remove concurrency key
-  presence.del(getHandlerConcurrencyKey(roomName));
+  await presence.del(getHandlerConcurrencyKey(roomName));
 
   // unsubscribe from remote connections
-  presence.unsubscribe(getRoomChannel(room.roomId));
+  await presence.unsubscribe(getRoomChannel(room.roomId));
```
hunkydoryrepair, line 588: these seem unnecessary.

packages/core/src/Room.ts
```diff
@@ -583,62 +587,73 @@ export abstract class Room<State extends object= any, Metadata= any> {
...
 
-    return reconnection;
+      const cleanup = () => {
+        delete this._reconnections[reconnectionToken];
+        delete this.reservedSeats[sessionId];
+        delete this.reservedSeatTimeouts[sessionId];
+        this._reconnectingSessionId.delete(sessionId);
+      };
+
+      reconnection
```
hunkydoryrepair, line 641: I've never been able to get the reconnection aspect to work well with colyseus so I'm not sure of the impact here, but it should still RETURN something. This has been change to have no return value, so the call cannot be awaited anymore

packages/core/src/Room.ts
```diff
@@ -796,7 +811,7 @@ export abstract class Room<State extends object= any, Metadata= any> {
     return await (userReturnData || Promise.resolve());
   }
 
-  private _onMessage(client: Client, bytes: number[]) {
+  private async _onMessage(client: Client, bytes: number[]) {
```
hunkydoryrepair, line 814: onMessage should not be awaited

hunkydoryrepair (commented):
There is a lot here that will break or slow down existing apps. 
Some beneficial changes, too, but I think a smaller PR with more isolated changes would be better.

PR #560, opened 2023-04-10
feat: make port optional in discovery node

Wenish:
Sometimes you dont want to specify a port so https and wss use the default port when the proxy connects to the game server

in cloud env the container in which the server runs may uses internally a other port then under which the server is externally avaiable.
common is internally 8080 and externally 443/80

so this change would allow to set the SELF_PORT env variable to an empty string "" and it would not override it with node.port
only if SELF_PORT is undefined or null

Example
```ts
const port = undefined ?? 3000 // port=3000
const port = null ?? 3000 // port=3000
const port = '' ?? 3000 // port=''
const port = '443' ?? 3000 // port=443
```

### DIFF
--- a/packages/core/src/discovery/index.ts
+++ b/packages/core/src/discovery/index.ts
@@ -13,7 +13,7 @@ async function getNodeAddress(node: Node) {
   const host = process.env.SELF_HOSTNAME || await ip.v4();
   const port = process.env.SELF_PORT ?? node.port;
   const baseAddress = `${node.processId}/${host}`
-  const shouldPortBeAddedToAddress = !!port;
+  const shouldPortBeAddedToAddress: boolean = !!port;
   return shouldPortBeAddedToAddress ? `${baseAddress}:${port}` : baseAddress;
 }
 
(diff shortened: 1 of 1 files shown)


packages/core/src/discovery/index.ts
```diff
@@ -11,8 +11,8 @@ export interface Node {
 
 async function getNodeAddress(node: Node) {
   const host = process.env.SELF_HOSTNAME || await ip.v4();
-  const port = process.env.SELF_PORT || node.port;
-  return `${node.processId}/${host}:${port}`;
+  const port = process.env.SELF_PORT ?? node.port;
+  return !!port ? `${node.processId}/${host}:${port}` : `${node.processId}/${host}`;
```
endel, line 15: To make it easier to read, I'd suggest writing this way:

```typescript
  const port = process.env.SELF_PORT ?? node.port;
  return (port !== undefined) 
    ? `${node.processId}/${host}:${port}` 
    : `${node.processId}/${host}`;
```

endel (changes requested):
Thanks for the PR @Wenish, it's a good addition - I'd suggest to avoid using `!!port` though, as it may be confusing to read. (I personally struggle with double negatives even on spoken language 😅 )

Wenish:
@endel what you think about how it is now?

Wenish:
done

Issue #489, opened 2022-01-08
RoomCountKey should be unique by deferent room class


amit-gshe:
Now if we use RedisPresence, all the room types will use a same room count key. Its ok when there is just one room type in the presence and the room will be created on a fewer rooms process. 

But when we have two types of processes and one type of processes for battle rooms and one another for lobby rooms. The lobby processes do not define a battle room. For now the lobby room will received the battle room creating request but the lobby room can not create a battle room so the request will be failed.

Its very nice that the matchmaker could customize the  room count key by deferent room type. Or a better support for muti-purposes room processes will be also fine.

hunkydoryrepair:
this really needs to be a callback so the application can handle the balancing logic.
There are going to be too many different scenarios to solve this generically.
For us, we really want to balance by players, but we may also want to separate by room type to support partitioning our very heavy load rooms onto dedicated servers. A lot of custom logic is really needed.

The callback pretty much needs the full matchmake data (full list of rooms w/ processId and metadata)

hunkydoryrepair:
`selectProcessIdToCreateRoom` exists now. I think this is done?

endel:
Closing as @hunkydoryrepair said, `selectProcessIdToCreateRoom` should cover the use case described on this issue. 🙌

Issue #466, opened 2021-10-28
Handling of joining concurrency


jldinh:
## Context

This follows a discussion started at https://github.com/colyseus/proxy/issues/14.

We observed that the colyseus-proxy sometimes gets a socket hang up error from the colyseus server.

In my experience, the socket hang up error occurs consistently when too many clients try to connect simultaneously (>50 clients started at the same time, especially when the state to sync initially is big, e.g., a few MBs). The colyseus server seems to start dropping connections then. I suspect this happens because the server's CPU becomes saturated, and the server cannot establish or maintain the websocket connections correctly.

I had a look at how the colyseus server handles connections to find the root cause, and was a bit confused by the use of a timer to spread out joining requests:
https://github.com/colyseus/colyseus/blob/85bacb8068adfc215eb53c12b1f0f681ae0bbd44/packages/core/src/MatchMaker.ts#L421-L449
What I understand is that new joining requests received before the previous ones have been processed are staggered by increments of 100 ms, but only up to a maximum of 2 s (with the default value). I believe the staggering is done to give the server some breathing room between requests to process other tasks, under the assumptions that one request takes less than 100 ms to process and that no more than 20 requests will ever be queued. If these assumptions are violated for an extended duration, the server will schedule more work than it can perform. I suspect this is the cause of the socket hang up errors.

Would it be better to use a queue system instead of the timer logic?

## Your Environment

* Colyseus Version: 0.14.20
* Node.js Version: 14.17.4
* Operating System and version: Ubuntu 20.04

jldinh:
bump

alexksso:
bump!

endel:
Assigned milestone 1.0 so this issue won't be closed again 🙈

PR #649, opened 2023-11-02
maintain correct room count

hunkydoryrepair:
Fixes this problem:

When a server is briefly unavailable (running any process more than 500ms), its roomcount is removed from redis.
When server finishes and closes a room, the room count goes negative.
This negative value turns that server into the "least busy" so all future room creations go to that one server, even if it actually has the most rooms (up to a point, anyway).

Solution:
Each server can trivially track its roomcount and set the count value, rather than rely on redis to INCR and DECR the value.

Other possible solutions:
Other solutions require either:
- Don't remove roomcount when server is inaccessible, which can lead to repeated failures when a server dies.
- Recount the number of rooms from Matchmaker Driver periodically
- Shutdown the server that didn't respond, either a self-initiated shutdown or forced externally.

### DIFF
--- a/packages/core/src/MatchMaker.ts
+++ b/packages/core/src/MatchMaker.ts
@@ -44,6 +44,8 @@ export let driver: MatchMakerDriver;
 export let isGracefullyShuttingDown: boolean;
 export let onReady: Deferred;
 
+export let roomCount = 0;
+
 export async function setup(
   _presence?: Presence,
   _driver?: MatchMakerDriver,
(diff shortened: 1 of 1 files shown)


endel:
Thank you so much for your PR and for sharing your thoughts on https://github.com/colyseus/colyseus/issues/155#issuecomment-1793000937 and https://github.com/colyseus/colyseus/issues/489#issuecomment-1792952871. These are all issues that relate with each other - plus the discussion at https://github.com/colyseus/colyseus/issues/466. 

I've just published `@colyseus/core` version `0.15.10` with your PR + @theweiweiway's https://github.com/colyseus/colyseus/pull/650 🙌

PR #650, opened 2023-11-03
remove redundant REMOTE_ROOM_SHORT_TIMEOUT and set it to default 2000

theweiweiway:
there were two REMOTE_ROOM_SHORT_TIMEOUT variables being exported, but only one was used.

I removed the redundant one and make the default value 2000 instead of 500, which is the value of the REMOTE_ROOM_SHORT_TIMEOUT that was NOT being used. Not sure which default value you were intending to use.


hunkydoryrepair:
can't you just set the env variable to change to 2000?

endel:
Thank you for the PR @theweiweiway.

In addition to your change, I've added a hardcoded `500` in the concurrency wait time to avoid long waits on concurrent `joinOrCreate` (https://github.com/colyseus/colyseus/commit/14de5227fa35b62edfb5dd9fdf6257559c4b06b2) - which desperately needs a proper solution in order to avoid these arbitrary timeouts that await for room creation. This problem has been described very well by @hunkydoryrepair here https://github.com/colyseus/colyseus/issues/155#issuecomment-1793000937 (Thanks again @hunkydoryrepair) 

---

(It turns out the entire `Utils.ts` file was duplicated in a previous refactor. Thank you for raising this!)

Issue #633, opened 2023-09-29
[Bug]: onLeave called before onJoin


hunkydoryrepair:
### Context

we have an async onAuth for the room, and an onLeave.

client disconnects before onAuth completes.

### Bug description

The onLeave is being called before onAuth even completes, which is well before onJoin is called.
This did not seem to be the case with 0.14, but is new with 0.15.

Not only that, but after onAuth completes, onJoin is called AFTER the onLeave.

It creates a number of issues that are pretty difficult to handle. There doesn't seem to be a publicly accessible flag on the client object to even determine this happened. I'm considering adding something to userData, but we would be required to check after every `await` in our onAuth and maybe even the `onJoin` 

Expectation is that onLeave would be called only after onJoin. If socket is disconnect during the onAuth/onJoin process, it should delay the call to onLeave until after these are complete. 

As it is, we create state for players that are not connected, and we also prevent them from logging in again.

### Steps to reproduce

1. use async onAsync
2. Disconnect client while onAsync is awaiting network call

### Environment & Versions

0.15

hunkydoryrepair:
I've workaround using userData on the client object, which works, but does require checking it in about 14 places and adding some code to reverse our database changes if we have to abort (normally, our onLeave handles it, but since that gets called BEFORE, and does not get called again later, we are forced to simulate the onLeave in some edge cases)

endel:
Thank you for reporting @hunkydoryrepair. 

This issue was introduced https://github.com/colyseus/colyseus/commit/5b69715e1796bce8a4374ae4f0dcd2b8c2ffcc65, now we have tests to cover this.

Please let me know if the latest version fixes it for you. Cheers!

hunkydoryrepair:
Amazing work, endel! Will test this.

hunkydoryrepair:
In previous tests, onLeave would also be called during the async onJoin sometimes (less often for us, as we do less async work in that case). Looking at the fix, I'm not sure it addresses onLeave while awaiting onJoin as well as the onAuth issue.

hunkydoryrepair:
This also changes the behavior of the `clients` array during onAuth.  I don't think it impacts us, but I can see that being a breaking change.

And, it looks like we can still get onLeave DURING onJoin now. So we have this problem:
 - in onJoin, record in our db that player is in game, verify using db they are not already in game elsewhere, then add to room state
 - onLeave received while updating db. Player not in room state, so onLeave does basically nothing.
[...]

PR #652, opened 2023-11-10
Distribute rooms based on player count or room count

hunkydoryrepair:
This is one step toward improving load balancing, which is an option to load based on player count.

 Number of players is likely to change much more rapidly than # of rooms, so this limits how frequently it makes REDIS updates to avoid flooding REDIS servers. 

it also maintains the "roomcount" usage as long as your balancing mode is the default. Thus, it should not break existing apps. However, switching to player load balancing will use a different key in REDIS.

This works in the server init, with a new server option:
```
  const server = new Server({
    greet: false,
    balanceMode: BalanceMode.players
    presence,
    driver
  });
 ```
 
 default remains room balancing. 
 
 There is also a small bug fix. `remoteRoomCall` was not using  the `rejectionTimeout` argument in a meaningful way. Needs to be passed to `requestFromIPC`


hunkydoryrepair:
This is actually not working right now in our tests because `MatchMaker.onClientJoinRoom` is not getting called

hunkydoryrepair:
I've updated this PR with a fix to #633 and a few other fixes to keep the `join` and `leave` events to be equal. There were a number of places they are wrong:

- exception during onJoin or onAuth would cause 'leave' event to be fired without a 'join' event
- a reconnection would cause an extra 'join' event to be fired.

Added some more tests for:
- Client disconnecting during async onJoin (onLeave should be called AFTER onJoin completes)
- onJoin throwing exception (should not call onLeave)

vitalyrotari:
You put to match attention on Redis optimization. Redis is the fastest in memory cache service, it can process millions of requests without any problems.  I build a game with vertical & horizontal scaling based on Colyseus + Redis, and I don't have any problems with performance or some bottlenecks, the number of players was 4-5k in same time, and the guaranty capacity was 10-12k of players. 

One more here, Redis driver is just an option, in case of Colyseus you can use another driver from the box or custom one. My opinion is to not garbage the core but improve it and make more customizable and flexible.

One of the bottlenecks on high loaded project will be the network bandwidth, and you must have separate load balancer which will distribute connections between clusters/servers.

endel:
Thanks for your input @vitalyrotari, the ability to distribute rooms can be really useful in case of having room definitions with different `maxClients` configurations - the default behaviour of selecting the process with least amount of rooms within the cluster could not be optimal in certain scenarios. https://github.com/colyseus/colyseus/pull/654 now addresses this allowing the end-user to provide an option to customize this behaviour 🙌 

Thank you so much for raising this @hunkydoryrepair, and for the initial implementation

PR #653, opened 2023-11-11
Don't Crash REDIS

hunkydoryrepair:
When roomcaches data gets quite large (600KB + for 2000 active rooms), the number of requests for the roomcache data can happen faster than the server is capable of READING the data. This causes the REDIS output buffers to grow and grow, until redis server eventually runs out of memory.

This change shares the same READ with multiple requests, so that each client is only requesting the roomcaches data 1 at a time. Once it has read/received the roomcaches data, it will proceed to get the updated version on the next call.

### DIFF
--- a/packages/drivers/redis-driver/src/index.ts
+++ b/packages/drivers/redis-driver/src/index.ts
@@ -72,8 +72,25 @@ export class RedisDriver implements MatchMakerDriver {
     }
   }
 
+  // gets recent room caches w/o making multiple simultaneous reads to REDIS
+  private _roomCachePromise: Promise<Record<string, string>> | null = null;
+  private getRoomCaches() {
+    if (this._roomCachePromise !== null) {
+      // give same list to all those who wait.
+      return this._roomCachePromise;
+    }
+    var newPromise = this._client.hgetall('roomcaches');
+    this._roomCachePromise = newPromise;
+    newPromise.then((result) => {
+      // clear shared promise so we can read it again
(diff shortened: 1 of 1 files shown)


hunkydoryrepair:
This will actually result in much faster joining of a room and MORE up to date room data. It might be counter intuitive, but REDIS doesn't queue up requests, it queues up RESPONSES. Those responses can sit in the queue for a while so by the time it is retrieved, it is out of date. By only requesting one at a time, the delay from the queue will be eliminated and room data will be as current as possible.

endel:
Thank you for the PR @hunkydoryrepair. I've made some improvements on top of your change (See https://github.com/colyseus/colyseus/commit/26f6ad4a580f39901131ed6acdb4d4a5458d75ff):

- Concurrent requests **per room name** will share the same `Promise`
- Each shared promise will first filter cached entries by room name before parsing the entry as JSON

Let me know if you have more feedback about this. Version `0.15.5` has been published!

hunkydoryrepair:
I'm not sure why you'd make separate requests for the same data just because the room name is different. Once the query changes to only look for the data by that name, it makes sense. For us...not too big a deal since we only have 2 room names that we use.

hunkydoryrepair:
I see you moved the filter into the shared "cache", although not really a cache as it doesn't store a copy of the data. 
I considered that approach but it seemed to be less useful. Saving only the filtered data to reuse requires making more queries to REDIS, which is what we are trying to avoid.

Issue #634, opened 2023-10-01
[Feature]: Callback for new room creation: create new rooms based on total players, not # of rooms.


hunkydoryrepair:
### Context & Description

Our "rooms" have very varied number of players, from 1 to 500. 
A single node.js instance can handle a couple thousand players, at least, but sometimes that is just a couple rooms.
While a server with more rooms but a lot fewer players is underutilized.

Tracking # of players per room, and spawning new rooms in server with fewest players would be a benefit to us.

A more generalized feature that exposes an API or callback that lets us apply our own room distribution logic.
For a room creation callback, controlling the target server, as well as being able to reject the room, or change the options on the room could be useful.

For example, we often have a mode where we shut down all rooms or certain rooms. There is no current way to prevent their creation, so the rooms get created and we just have to deny entry to everybody. That works fine, but it means we have a room instance running that we don't really want.

We also have no way to validate the filters. We have 100,000 valid options for the room filters, but no way to detect a bogus room creation until it is already being created. 

### Use cases

goals and philosophy of the project...This is a heavy question, but I think contirubting to the smooth operation of a multiplayer game platform would apply to a lot of projects. Any project where the rooms can vary with # of players greatly could benefit.

### Proposed API

Have not thought this through, but a callback on the server for onRoomCreate(), perhaps, the exposes the list of servers and the available room information, and can return with a definition of where to create it, and probably an opportunity to thrown an error or prevent room creation.

`server.onRoomCreate = (serverInfo) => roomDisposition`

hunkydoryrepair:
I have a branch that gives the option during Colyseus setup to use player count (CCU) vs room count in balancing.
I can do a PR if we might want to just have that, since doing a callback to allow that seems pretty challenging. 
I set it up to only update player counts once per second when using players to balance load.

endel:
Hi @hunkydoryrepair, thank you for sharing the detailed use-case. This would be really great to have - if you can share I'd love to see what you envision for this!

endel:
Addressed on https://github.com/colyseus/colyseus/pull/654 🙌

PR #654, opened 2023-11-14
Room distribution / Allow to customize which process creates the room

endel:
This is a draft PR that is meant to cover @hunkydoryrepair's proposal from #652 and #634. 

---

Includes changes:

- Introduced an internal `Stats` module, responsible for handling the `roomcount` key on Redis.
- The `roomcount` hash now holds its values separated by comma (`processId`->`"count,ccu"`) 
- The `Stats` module is exposed via `matchMaker.stats`, so end-user can call:
  - `matchMaker.stats.fetchAll()` -> `Promise<Array<{ roomCount: number; ccu: number; processId: string }>>`
  - `matchMaker.stats.getGlobalCCU()` -> `Promise<number>`
- Add `selectProcessIdToCreateRoom` option to `Server` constructor. 

The default value for `selectProcessIdToCreateRoom` behaves exactly as previous versions, but consuming the new `roomcount` structure:

```typescript
selectProcessIdToCreateRoom = async function (roomName: string, clientOptions: any) {
    return (await stats.fetchAll())
      .sort((p1, p2) => p1.roomCount > p2.roomCount ? 1 : -1)[0]
      .processId;
  }
```

---

**_EDIT: Renamed `assignRoomToProcessId ` to `selectProcessIdToCreateRoom`_**


hunkydoryrepair:
This is great, and will be good to try to customize them.

But the ccu and player counts will be way off.

Please look at the tests I added in my PR. Those will expose some of the problems.

The player count and ccu quickly go negative, because `leave` event is currently fired way more often than the `join` event, and you get an extra `join` event during reconnection. My PR addresses those issues, and the problem with `onLeave` being called during onJoin or onAuth. The stats are useless if they are inaccurate.

endel:
Thanks @hunkydoryrepair, I'll check it!

endel:
I didn't realised you had tests for this and a proposal for fixing the onJoin/onLeave on #652, thank you for the test cases - I've just brought them to this PR, and took a slight different approach for the onJoin/onLeave you reported on #633.

I'm going to add some more tests to cover async onJoin/onLeave + reconnection agasint the new `matchMaker.stats` API, then I think this is ready! 🙌

endel:
All seems good here now! 

The final method name for the "room distribution" method so far is **_`selectProcessIdToCreateRoom`_**. 

@hunkydoryrepair do you think we're missing something else here? 🙏

hunkydoryrepair:
nice. this looks like it will work well to me. I had considered something like that approach for onLeave but I didn't understand why it had changed from that technique earlier, so I tried to do something with the least impact, especially since we don't do reconnects and have no way to test that. But just changing client state and calling onLeave later...perfect!
It isn't clear that the matchmaker events are changed, however, or if you changed that completely. I will look closer, but the 'join' and 'leave' events emitter to matchmaker was independent of the room.onJoin/room.onLeave calls.
This looks like we can test with it and try on production next week. Under our load, we will see if player counts go wrong very quickly, so will keep it with room count until we verify player counts stay good.

hunkydoryrepair:
ok, pretty sure this will still call this:

      this._events.emit('leave', client, willDispose);
      
     without a corresponding 
     
     
      this._events.emit('join', client);

I will point out in a review

packages/core/src/Room.ts
```diff
@@ -599,18 +614,23 @@ export abstract class Room<State extends object= any, Metadata= any> {
...
-    this._events.emit('join', client);
+    // emit 'join' to room handler (if not reconnecting)
+    if (!previousReconnectionToken) {
+      this._events.emit('join', client);
+    }
 
-    // allow client to send messages after onJoin has succeeded.
-    client.ref.on('message', this._onMessage.bind(this, client));
+    // state might already be ClientState.LEAVING here
+    if (client.state === ClientState.JOINING) {
```
hunkydoryrepair, line 623: does this work during RECONNECT? It looked to me like the client.state would be RECONNECTING, and the code below needs to run when reconnecting as well as when JOINING

packages/core/src/Room.ts
```diff
@@ -583,9 +583,24 @@ export abstract class Room<State extends object= any, Metadata= any> {
         if (this.onJoin) {
           await this.onJoin(client, options, client.auth);
         }
+
+        client.ref.removeListener('close', client.ref['onleave']);
+
+        // client left during `onJoin`, call _onLeave immediately.
+        if (client.state === ClientState.LEAVING) {
+          await this._onLeave(client, Protocol.WS_CLOSE_GOING_AWAY);
```
hunkydoryrepair, line 591: by calling _onLeave here, the _onLeave function will emit `leave` event to matchmaker before the 'join'
The net should be correct, since the 'join' is still sent below, but emitting these out of order seems potentially problematic.

packages/core/src/Room.ts
```diff
@@ -599,18 +614,23 @@ export abstract class Room<State extends object= any, Metadata= any> {
       }
     }
 
-    // emit 'join' to room handler
-    this._events.emit('join', client);
+    // emit 'join' to room handler (if not reconnecting)
+    if (!previousReconnectionToken) {
+      this._events.emit('join', client);
```
hunkydoryrepair, line 619: here, if player has disconnected, the 'join' is being emitted after 'leave'

packages/core/src/Room.ts
hunkydoryrepair: if state before calling is changed to JOINING, however, it should still work.

packages/core/src/Room.ts
hunkydoryrepair: since this is only for a new join, anyway, could put it right above where it checks for LEAVING

packages/core/src/Room.ts
endel: Yes! In fact the `ClientState.RECONNECTED` is only used to prevent the `'leave'` event from being emitted on the first connection after the reconnection has been confirmed. https://github.com/colyseus/colyseus/blob/4ff32c4fe2b143538ac75a6823455c5835f135a5/packages/core/src/Room.ts#L940-L945

packages/core/src/Room.ts
endel: Well observed! It turns out it's safe to move the `'join'` event before this call, as it is already being explicitly ignored if `!previousReconnectionToken`. So that's the only code path the `'join'` event could happen 🙌

packages/core/src/Room.ts
endel: You're right! Thanks for spotting this!

packages/core/src/Room.ts
hunkydoryrepair: Sorry, I meant when reconnecting, the state may be LEAVING or JOINED, because it never gets set back to JOINING after it gets set to JOINED.
I think I explicitly set it to JOINING in beginning of that function, but I didn't see if you copied that.

PR #656, opened 2023-11-27
Change setPatchRate to accept milliseconds as null

rafaeor:
In the documentation link provided (https://docs.colyseus.io/server/room/#broadcastpatch), you can disable PatchRate by using this.setPatchRate(null). 
However, I encountered an error while attempting to set setPatchRate to null. The error message states: 'Argument of type 'null' is not assignable to parameter of type 'number'.' To resolve this issue, I suggest modifying the code as follows to allow null as an argument as well.

### DIFF
--- a/packages/core/src/Room.ts
+++ b/packages/core/src/Room.ts
@@ -323,7 +323,7 @@ export abstract class Room<State extends object= any, Metadata= any> {
     }
   }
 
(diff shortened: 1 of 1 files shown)


endel:
Thanks a lot @rafaeor 🎉 Just released `@colyseus/core` version `0.15.15` with your fix

Issue #570, opened 2023-05-07
simulateLatency doesn't go back 


endel:
Reported by Cucumba on Discord

---

Server latency simulation bug as previously mentioned here ⁠🐛│bugs⁠
https://github.com/colyseus/colyseus/blob/0.15/packages/core/src/Server.ts#L239
https://github.com/colyseus/colyseus/blob/0.15/packages/transport/ws-transport/src/WebSocketTransport.ts#L73
https://github.com/colyseus/colyseus/blob/0.15/packages/transport/uwebsockets-transport/src/uWebSocketsTransport.ts#L124
Reason: previous function is being overridden each time when simulateLatency is called, it sets timeout for existing timeout, instead of original function

Here is example:

```typescript
class Logger {
    log(message) {
        console.log(message)
    }
}

function simulateLatency(milliseconds) {
    const previousLog = Logger.prototype.log
    Logger.prototype.log = function(...args) {
        console.warn(milliseconds)
        setTimeout(() => previousLog.apply(this, args), milliseconds);
    }
}

const logger = new Logger()
logger.log("msg") // no delay

simulateLatency(1000)
logger.log("msg") // delay 1 second

simulateLatency(0)
logger.log("msg") // delay 1 second

simulateLatency(1000)
simulateLatency(1000)
simulateLatency(1000)
logger.log("msg") // delay 4 seconds
```

afrokick:
I fixed it in #663 + some test cases have been added

endel:
Thank you @afrokick 🎉

PR #663, opened 2023-12-23
Fix simulateLatency

afrokick:
- [x] add test cases
- [x] fix behaviour
- [x] disable latency when passing 0

Related issue #570


endel:
Thank you so much for this, I'll merge it after the holidays 🎅  Merry Christmas @afrokick

PR #668, opened 2023-12-25
[LobbyRoom] Update listed rooms on the client after rooms are marked as private

### LINKED ISSUE #617: [Bug]: Changing room to private doesn't update realtime listing
### Context

I'm trying to use realtime listing on my game and i want to allow player to set a room to be private or public after joining the room. 

### Bug description

When a room is changed to public, all the clients connected to the lobby will get `+` message and the listing is updated normally. But when changing it back to private, there is no message at all to the clients so the listing doesn't get updated. Calling `updateLobby` also doesn't work.

The current workaround is:
```js
this.setPrivate(newPrivate).then(() => {
	this.presence.publish("$lobby", `${this.roomId},${newPrivate ? 1 : 0}`); // 1 = removed, 0 = added
});
```

### Reproduction

_No response_

### Steps to reproduce

1. Create a server with lobby room
2. Connect a client to lobby
3. Create a public room in that lobby
4. Set that room to private 

### Environment & Versions

Colyseus Version: 0.15.5
Node.js Version: 18.16.0
Operating System: Windows 10

afrokick:
Fix #617

Proposed solution:

There are 2 new room's events: `setPrivate` and `setPublic`.

I've also changed the `updateLobby` logic: the remove message always propagated even for private rooms.

### DIFF
--- a/bundles/colyseus/test/rooms/LobbyRoomIntegration.test.ts
+++ b/bundles/colyseus/test/rooms/LobbyRoomIntegration.test.ts
@@ -1,12 +1,7 @@
 import assert from "assert";
-import sinon from "sinon";
 import * as Colyseus from "colyseus.js";
-
-import { matchMaker, Room, Client, Server, ErrorCode } from "../../src";
-import { DummyRoom, DRIVERS, timeout, Room3Clients, PRESENCE_IMPLEMENTATIONS, Room2Clients, Room2ClientsExplicitLock } from "./../utils";
-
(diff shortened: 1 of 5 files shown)


endel:
Thank you so much for this @afrokick 🎉  
I just refactored a bit to use a `"visibility-change"` event instead of `"setPrivate"` + `"setPublic"` 🙌

afrokick:
@endel Thanks! Agree, visibility-change is cool!

Issue #673, opened 2024-01-03
[Feature]: validate Room Options/Availability


hunkydoryrepair:
### Context & Description

Our game design requires that we make certain rooms, or all rooms, unavailable to certain players or at certain times.
We also have a lot of bots and hackers attempting to send invalid data.

We only use a small number of room types, because there are over 200,000 different valid room options. We cannot define/register that many room types. However, we need to ensure we don't create invalid rooms. We also want to prevent duplicate rooms and prevent creating rooms when we are doing temporary maintenance.

PROBLEM: There isn't a good mechanism for denying a room creation or returning an error to the caller (client, typically) for creating the room. If we throw an error during `onCreate`, which is what we currently do, the SERVER is removed! The server is treated as dead and removed from roomcount on REDIS!

Workarounds for this are a bit difficult. If we cannot get an error back to the caller the clients will run into a different set of errors. Instead of throwing errors, we could create a "disabled" room, set it to private, and immediate destroy it. 

### Use cases

Limiting times certain rooms are available
Allowing server maintenance while process is running.
Allowing only valid room configurations to be created.
Allowing singularity rooms (rooms with only a single instance allowed).

### Proposed API

Simplest solution: don't invalidate the server just because it throws an error during `onCreate`! This seems unnecessary, and should only happen if it is a timeout. I think if there is an error thrown creating a room on any server, expect it to happen on all servers and return the error to the caller.

Alternative solution: onPreCreate to check the viability of creating the room, similar to onAuth before onJoin. Problem with this approach is it doesn't support the singularity rooms, or can cause other timing/synchronization issues. 

Being able to reject room creation within onCreate (maybe a specific exception type?) is really pretty much required.

hunkydoryrepair:
This seems to be a new issue we encountered. I think the latest release is more aggressive at killing servers, which causes some serious problems when the servers have no active users, because they don't recover quickly at all to restore their entry into `roomcount`

endel:
Hi @hunkydoryrepair, thanks for the detailed report as always - I think we've been missing a very obvious scenario that you just described:

If the "create room" request has been forwarded to a different process, and an error is thrown on that process, it is currently being treated as if the process fails to respond (`"ipc_timeout"`), and the "create room" request is retried at the original node. I just created a pull-request to only perform the retry in the specific case of `"ipc_timeout"` - other errors types are re-thrown: https://github.com/colyseus/colyseus/pull/674

[...]

PR #674, opened 2024-01-03
only re-tries 'create room' on 'ipc_timeout' errors. re-throw otherwise.

endel:
Fixes https://github.com/colyseus/colyseus/issues/673

### DIFF
--- a/bundles/colyseus/test/IPC.test.ts
+++ b/bundles/colyseus/test/IPC.test.ts
@@ -40,6 +40,31 @@ describe("Inter-process Communication", () => {
         });
       });
 
+      it("IPC timeout error should be thrown", async () => {
+        try {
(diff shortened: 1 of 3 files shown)


hunkydoryrepair:
yes! This was exactly the code that was causing us problems. I was thinking it always used IPC for creating the room, even if on the same server. Does it not?

endel:
Thank you for reviewing - just released `@colyseus/core` v`0.15.19` with this.

If the selected process id is the current process it does a direct call without IPC:

https://github.com/colyseus/colyseus/blob/32c7bd04d2a85854a4a4429f9dd80d62f8810599/packages/core/src/MatchMaker.ts#L362-L367

Otherwise, it goes through the IPC

PR #692, opened 2024-02-04
@colyseus/auth: handle errors in post /anonymous route

alexeychikk:
In my use-case I'm validating `options` in `onRegisterAnonymously` callback and throwing a error if it is invalid, but the app crashes.
So I added some error handling for that.


endel:
Thank you for the contribution @alexeychikk 💙 Just released `@colyseus/auth` version `0.15.10` with your fix

PR #706, opened 2024-03-20
support authentication module on TestServer

jblemee:


packages/testing/src/TestServer.ts
```diff
@@ -10,6 +10,9 @@ export class ColyseusTestServer {
     create: Client['create'],
     joinById: Client['joinById'],
     reconnect: Client['reconnect'],
+    registerWithEmailAndPassword: Client['auth']["registerWithEmailAndPassword"];
+    signOut: Client['auth']["signOut"];
+    signInWithEmailAndPassword: Client['auth']["signInWithEmailAndPassword"];
```
endel, line 15: Can we have the entire `Client['auth']` reference here under `auth` instead of binding the individual `auth.*` methods?

endel:
Hi @jblemee, that's a great addition, thank you for your PR! I've added just a single comment in the review, will merge this soon!

packages/testing/src/TestServer.ts
jblemee: @endel Done ;)

endel:
Thank you so much @jblemee 💙 Just released `@colyseus/testing@0.15.3` with your changes 🎉

PR #711, opened 2024-03-30
delete unused Travis config

kevinji:
I'm assuming all builds are using GitHub Actions rather than Travis now.


endel:
Thank you @kevinji

PR #712, opened 2024-03-30
set up pnpm workspaces with Lerna

kevinji:
Setup is based on https://lerna.js.org/docs/recipes/using-pnpm-with-lerna

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,15 +1,9 @@
 {
-  "packages": [
-    "bundles/*",
-    "packages/*",
-    "packages/drivers/*",
-    "packages/transport/*",
-    "packages/presence/*",
-    "packages/serializer/*"
-  ],
   "version": "independent",
   "publish": {
     "message": "chore(release): publish"
   },
-  "hoist": true
+  "hoist": true,
+  "useWorkspaces": true,
+  "npmClient": "pnpm"
 }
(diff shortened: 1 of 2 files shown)


endel:
That's some nice clean-up, thank you @kevinji 💙

endel:
I just noticed the `lerna run test` is failing now:

```
❯ pnpm test

> @colyseus/monorepo@0.15.0 test /Users/endel/Projects/colyseus/colyseus
> lerna run test --stream

lerna notice cli v7.4.2
lerna ERR! ECONFIGWORKSPACES The "useWorkspaces" option has been removed. By default lerna will resolve your packages using your package manager's workspaces configuration. Alternatively, you can manually provide a list of package globs to be used instead via the "packages" option in lerna.json.
 ELIFECYCLE  Test failed. See above for more details.
```

Lerna references here are mostly legacy, I'd love to remove `lerna` completely and use `pnpm` wherever is possible. In the past we used `lerna` for publishing isolated packages - but `pnpm` already does this quite well. Now the only references to `lerna` are in that test script and in the `build` script: https://github.com/colyseus/colyseus/blob/019f446b13e048b4c139abc336ba87f860da6e7f/build.ts#L8-L9

kevinji:
Hey! Sorry about that, I think their instructions about pnpm integration are slightly out of date. I just opened #713 to remove that field.

endel:
That's great! Thank you @kevinji ☺️

PR #721, opened 2024-04-24
[bun-transport] fix parsing body string to JSON in Matchmaker

afrokick:
onCreate/onJoin  `options` parameter's type is string now, but must be an object.

The problem was in matchmaker. It doesn't convert string body to json object.


endel:
Thanks a lot for this @afrokick. I have the impression `req.body` was available as a parsed JSON in previous versions, do you know if this behaviour has changed recently in Bun? I'll merge this shortly!

afrokick:
The doc says that they reimplemented `Bun.serve`.
https://bun.sh/docs/api/http

PR #724, opened 2024-04-29
feat: support asBytes option in room.broadcast

tyzoo:
Colyseus currently supports `client.sendBytes()` to send messages with raw bytes, though it is lacking a comparable feature for broadcasting raw bytes to all clients. This proposal is to support `room.broadcastBytes()` and/or `room.broadcast(,,{ asBytes: true })`.


hunkydoryrepair:
This definitely seems like a missing feature. We could use this.

endel:
This PR has been merged in `0.16` branch - it's going to be closed once stable version is out  🙌

endel:
Version 0.16 introduced a `broadcastBytes()` method. Thank you for the PR @tyzoo

PR #737, opened 2024-06-09
add error handling for schema corruption

hunkydoryrepair:
There are places in colyseus where exceptions can be thrown, usually do to application errors, but cannot be handled by the application. These unhandled exceptions result in the node server shutting down, and usually ungracefully so the room onDispose methods and the player onLeave calls are not made.

To prevent common cases, colyseus can catch the schema serialization errors and shut down the room, leaving other rooms on the same server to continue running, and save a lot of disruption. 

These errors can frequently be result of deliberate hacking attempts by people looking to cheat. Crashing a single room is preferable to a non-graceful shutdown of the entire application.

### DIFF
--- a/packages/core/src/Room.ts
+++ b/packages/core/src/Room.ts
@@ -495,7 +495,16 @@ export abstract class Room<State extends object= any, Metadata= any> {
       return false;
     }
 
-    const hasChanges = this._serializer.applyPatches(this.clients, this.state);
+    let hasChanges = false;
+    try {
+      hasChanges = this._serializer.applyPatches(this.clients, this.state);
+    } catch (e) {
+      debugAndPrintError(e);
+      // corrupt state, cannot recover and will impact all clients.
+      // this room needs to shut down.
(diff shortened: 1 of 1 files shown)


endel:
Schemas v3 has been refactored recently to avoid triggering errors at encoding time. Now instance validation run during assignment instead: https://github.com/colyseus/schema/blob/9107fe8fee7a119dfa74cec8d71325df5917c096/src/annotations.ts#L393-L398

e.g.: this will throw an error:

```typescript
class State extends Schema {
  @type([item]) items = new ArraySchema<Item>();
}

const state = new State();
state.items.push(new Player()) // wrong instance type, will throw an error
```

(Going to close this issue once 0.16 is released as stable... 🙏 )

endel:
Colyseus 0.16 no longer throws errors at encoding time (unless it's an actual unknown encoding bug rather than user data type error)

We also have a [custom error handling within rooms](https://docs.colyseus.io/server/room/exception-handling) 🎉

PR #751, opened 2024-07-23
loadtest: Override console.debug as well

francislavoie:
As mentioned on Discord, I noticed that `console.debug` wasn't being overriden, so when my app was printing to debug (in a lot of places) it would write on top of the `blessed` UI.

### DIFF
--- a/packages/loadtest/src/index.ts
+++ b/packages/loadtest/src/index.ts
@@ -247,11 +247,16 @@ Example:
     screen.append(networkingBox);
     screen.render();
 
+    const debug = console.debug;
     const log = console.log;
     const warn = console.warn;
     const info = console.info;
     const error = console.error;
 
+    console.debug = function(...args) {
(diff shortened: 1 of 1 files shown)


endel:
Thank you so much for the addition @francislavoie 🎉  Just released new version with it!

endel:
Oops, sorry I missed that one... Just published again... Thank you!! 😊

francislavoie:
I pulled down the changes, can confirm it fixes writing the text inside the UI box. But it doesn't use the color I expected it to, it still prints it in white. Got a clue why that might not be working? If I do like `{#777-fg}` it works though, but that wouldn't use the user's terminal color for grey. Weird.

As an aside, it looks like `blessed` is super abandoned tho (9 years since last commit, yikes). There's a relatively recent fork https://github.com/kenan238/reblessed, or https://github.com/vadimdemedes/ink for a totally different modern approach. Maybe worth looking into.

Also (I can post this in #752 too) it still doesn't seem to be tracking networking bytes & connection count for some reason.

PR #752, opened 2024-07-23
loadtest: Also wrap create & joinById for tracking stats

francislavoie:
Currently only `joinOrCreate` and `join` are being wrapped, but I typically use `create` and `joinById` in my app. Those also need to be wrapped the same way, I think.

### DIFF
--- a/packages/loadtest/src/index.ts
+++ b/packages/loadtest/src/index.ts
@@ -460,9 +460,23 @@ Example:
         return room;
     }
 
+    const _originalCreate = Client.prototype.create;
+    Client.prototype.create = async function(this: Client) {
+        const room = await _originalCreate.apply(this, arguments);
+        handleClientJoin(room);
+        return room;
+    }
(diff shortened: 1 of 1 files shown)


francislavoie:
I'm not sure what the deal is, but it seems like the wrapper functions don't get called for me. I put in a `console.log` inside of `handleClientJoin` and in each `prototype` wrapper and it just doesn't get logged. Really weird. Is there somekind of weird order of operations thing with ES modules making it not take effect?

francislavoie:
It seems like `import { cli, Options } from "@colyseus/loadtest";` pulls in the `build/index.js` and not the `build/index.mjs` for some reason. That might be an issue with my project config though. So I think the loadtest overrides get applied to the commonjs version of Client instead of the esm one. What am I doing wrong :thinking:

francislavoie:
Ah I figured it out, PR coming

PR #753, opened 2024-07-23
loadtest: Modernize package.json to properly support ESM

francislavoie:
See comments in https://github.com/colyseus/colyseus/pull/752#issuecomment-2245880073

Since `type` & `exports` were missing, my project was importing the CommonJS build instead of the ESM one, using Node v22. Adding these in makes the `Client.prototype` stuff work correctly because it overrides the same functions that my project is actually using.

### DIFF
--- a/packages/loadtest/package.json
+++ b/packages/loadtest/package.json
@@ -2,6 +2,14 @@
   "name": "@colyseus/loadtest",
   "version": "0.15.5",
   "description": "Utility tool for load testing Colyseus.",
+  "type": "module",
+  "exports": {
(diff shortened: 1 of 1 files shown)


endel:
Thank you again @francislavoie 🎉  

(ESM support is still missing on most modules on 0.15.x... Last week I've added ESM support on 0.16 here https://github.com/colyseus/colyseus/commit/7e2df366778391443ea52dfe9f96fa21ebd36623 - the 0.16 version is still under "preview" tag on NPM)

PR #758, opened 2024-08-04
bump: @colyseus/uwebsockets-transport to support node.js 22

jetkai:
Bumps the **uWebSockets.js** dep from v20.43.0 to [v20.44.0](https://github.com/uNetworking/uWebSockets.js/releases/tag/v20.44.0) inside the `@colyseus/uwebsockets-transport` package.

Resolves the following error when trying to run colyseus in Node 22:
```console
Uncaught Error: This version of uWS.js supports only Node.js LTS versions 16, 18 and 20 on (glibc) Linux, macOS and Windows, on Tier 1 platforms (https://github.com/nodejs/node/blob/master/BUILDING.md#platform-list).
```

### DIFF
--- a/packages/transport/uwebsockets-transport/package.json
+++ b/packages/transport/uwebsockets-transport/package.json
@@ -1,13 +1,13 @@
 {
   "name": "@colyseus/uwebsockets-transport",
-  "version": "0.15.7",
+  "version": "0.15.8",
   "input": "./src/index.ts",
   "main": "./build/index.js",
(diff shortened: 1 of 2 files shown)


endel:
Thank you for the PR @jetkai. 

FYI the `"overrides"` field on `package.json` can be used to force the version of a nested package: https://docs.npmjs.com/cli/v10/configuring-npm/package-json#overrides

PR #765, opened 2024-08-29
Add callback before gracefully shutdown server

colus001:
### WHY
Currently the server is unable to exclude process safely on shutdown. In order NOT to break current gracefully shutdown process I think it needs another shutdown callback to do something before "official" shutdown process.

Here's use case.

```
server.onBeforeShutdown(async () => {
  // Exclude self from match maker
  await matchMaker.stats.excludeProcess(matchMaker.processId);

  // wait until room count goes to 0 (let players continue to play)
  while (matchMaker.stats.local.roomCount > 0) {
    await new Promise((resolve) => setTimeout(resolve, 100));
  }
})
```

### WHAT
Simple added `onBeforeShutdown` lifecycle callback

---

IMO `await this.onShutdownCallback();` must called before `await matchMaker.gracefullyShutdown();`. If match maker removes all rooms and presence, there's almost nothing to process about the rooms.

### DIFF
--- a/packages/core/src/Server.ts
+++ b/packages/core/src/Server.ts
@@ -248,6 +248,7 @@ export class Server {
     });
 
     try {
+      await this.onBeforeShutdownCallback();
       await matchMaker.gracefullyShutdown();
       this.transport.shutdown();
       this.presence.shutdown();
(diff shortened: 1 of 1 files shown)


endel:
Hi @colus001, I agree with you, I've just published a new version of `@colyseus/core` (`0.15.49`) moving `await this.onShutdownCallback()`  before `await matchMaker.gracefullyShutdown()` 🙌  

Let me know if that works for you at the moment!

PR #772, opened 2024-09-22
Restore package.json for CJS support.

damian-pastorini:
Revert change from: https://github.com/colyseus/colyseus/pull/753/files Since this fully breaks CJS support.
Considering main is set to be the CJS entry: 
"main": "./build/index.js",
Then the package type must remain as commonjs otherwise you can't "require" it.


endel:
Thank you @damian-pastorini! Just released `@colyseus/loadtest` version `0.15.7` with this

PR #773, opened 2024-10-01
Include history logs on saved file

damian-pastorini:


francislavoie (approved):
For what it's worth, I approve. I was meaning to propose this change myself, was surprised to see the log not having the collected `console.log`

endel:
Thank you so much for this! Sorry for the delay to merge

PR #774, opened 2024-10-03
add bodyParser.json JsonOptions to ConfigOptions

Snirozu:
this should make people be able to set the limit of the received JSON data, etc.
(I haven't tested if this pr works, but i think it should)

### DIFF
--- a/packages/core/src/Server.ts
+++ b/packages/core/src/Server.ts
@@ -19,6 +19,8 @@ import { Transport } from './Transport';
 import { logger, setLogger } from './Logger';
 import { setDevMode, isDevMode } from './utils/DevMode';
 
+import { OptionsJson } from "body-parser";
+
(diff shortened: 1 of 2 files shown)


endel:
Hi @Snirozu, thanks for the PR. Perhaps we should specify some sane defaults to prevent abuse by default. 

(Not too long ago the `"maxPayloadLength"` option from the WebSocket transport has been changed to accept 4kb  max by default. https://github.com/colyseus/colyseus/commit/c995d544e3bf60a3d765fbf39b5212190cc2c601)

Snirozu:
decreased from the default value '100kb' to '5kb', i think it's an average size for a json data while not breaking most requests

Snirozu:
late bump but it would be really useful if it was implemented

PR #780, opened 2024-10-09
loadtest: Add clientId to identify each client in logs

francislavoie:
In my logs, I want to be able to tell apart which client replica is which, and currently there's no way to do so. Simple thing is to simply pass the index in the `options` which is available to `main()`.


endel:
Thank you @francislavoie 🎉 , sorry for the delay to merge

PR #791, opened 2024-11-07
add onUncaughtException() method for exception handling. #623

endel:
Closes https://github.com/colyseus/colyseus/pull/623

---

This PR introduces the ability to automatically catch unhandled exceptions from rooms userspace:

- **opt-out by default**. Landing this feature has **no effect** on anybody's apps & codebase.
- To opt-in, you should simply define your own `onUncaughtException(err, methodName, args)` in your room
- If the `room.onUncaughtException` is defined, all callbacks are going to be wrapped by try/catch: 
  - Lifecycle hooks: `onAuth`, `onJoin`, `onLeave`, `onCreate`, `onDispose`
  - On message
  - `this.clock.setInterval()` / `this.clock.setTimeout()`
  - `this.setSimulationInterval()`
- Even though exceptions are caught automatically, some are going to be re-thrown, due to how they're handled internally by the framework so the behaviour to keep connections active or not remain the same, such as: `onCreate`, `onAuth`, `onJoin`, `onLeave`.

New API looks like this:

```typescript
class MyRoom extends Room {
  // ..
  onUncaughtException(error: Error, methodName: string) {
    /**
     * by having this method defined, all exceptions wihin the game room will be "catched" here.
     */
    console.log(error, methodName);
  }
  // ..
}
```

Feedback is very welcome!


CookedApps:
@endel That's a great addition to the framework, looks solid. :+1: 

We used to have problems with uncaught errors in the past (_not caused by Colyseus_), which led to many server crashes. This affected the experience of all players on the instance, regardless of the severity of the error. We wanted to prevent that at all costs, so we built a custom catch-all solution similar to this. The need is definitely there. 

I wonder why not make that opt-in by default? What would be the implications?

MCArth:
Similar to Sandro we catch all errors. Seems like a good change :) Only thing that I might question would be the rethrowing -> could this still lead to process crashing? But wanting the behaviour of "Error thrown in onAuth -> kick client" 100% makes sense

heilmela:
@endle Thanks for picking up the issue again! 
I had a look at your changes and they definetly solve the issue I had in mind 👍  

One small suggestion that might make things even better - what do you think about adding some custom error types or narrow down methodName's type ?  

Either `typeof(Error) = OnJoinError | OnAuthError` etc.
or `methodName :  "onJoin" | "onAuth"` instead of string.

Having specific error types makes the API more robust in two ways - it makes the consumption of the easier, and it ensures API changes are caught at compile-time rather than silently failing when the signature changes (e.g. a new hook is added and the default case picks up a new error).

Endle:
> @Endle Thanks for picking up the issue again!

Hi heimela, there is a typo. You should reach to @endel  :)

heilmela:
Oops, my apologies

endel:
Thank you for the feedback! ❤️ 

@CookedApps when we reach version 1.0 we can make it opt-in by default! Right now I just don't want to change anybody's server behaviour when upgrading...

@heilmela That makes sense! I've refactored it a bit so each individual error type contains the relevant context data (`OnJoinException` has a `client` and `options` properties. `OnMessageException` has `client` and `payload` properties, etc.)

When overriding the `onUncaughtException` from base class unfortunately TypeScript does not keep the original types, so we need to provide them ourselves. If types are not provided, TypeScript uses `any` by default - I think there is an open issue about this on the TypeScript repository but I couldn't find it. Ideally both the `error` and `methodName` types should be inferred, but it's fine for now I believe... Hopefully TypeScript improves this in the future.

```diff
-onUncaughtException(error, methodName) { // can't do this 
+onUncaughtException(error: Error, methodName: string) { // need to provide the types again 
```

I'm planning to merge this and release new version today, let me know if you have any further feedback please 🙏☺️

PR #824, opened 2025-02-24
__dirname compatibility & .js extension fix

nmime:


nmime:
@endel can you also publish @colyseus/core (becaue of extensions)?

PR #826, opened 2025-02-24
allow for manually passing in a sessionId during the reserveSeatFor h…

raymondtri:
…ook via the authData as authData should be verified by that point. This allows for users to manually control the sessionId via the onAuth hook which adds lots of flexibility to transport methods.


raymondtri:
I also added a clientState.CLOSED enum for wider transport support

endel:
That's a great addition, thank you @raymondtri 🥳  Just released `@colyseus/core@0.16.10` with it

PR #828, opened 2025-02-26
Fix simulateLatency when using uWebSocketsTransport

rogino:
When using latency simulation with uWebSocketsTransport, I get the following error:

> Error: Text and data can only be passed by String, ArrayBuffer or TypedArray.

Using `Array.from(buf)` would convert it to an array which uWebSockets does not accept.


endel:
Thank you so much for the PR @rogino, just published `@colyseus/uwebsockets-transport@0.16.5` with your fix!

PR #849, opened 2025-04-06
expose Presence setMaxListeners

sylvainpolletvillard:
Node EventEmitter used in Presence has a built-in maxListeners used to try to detect and prevent memory leaks. Its default value is 10 listeners, which is short when you have more than 10 rooms that need:

>  (node:85265) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 server-announcement listeners added to [EventEmitter]. MaxListeners is 10. Use emitter.setMaxListeners() to increase limit

Problem is the emitter is not exposed by Presence, so there's no way to call this setMaxListeners method from colyseus.

This PR exposes this setMaxListeners method so that we can set the presence maxListeners without changing the default value for all EventEmitters.

### DIFF
--- a/packages/core/src/presence/LocalPresence.ts
+++ b/packages/core/src/presence/LocalPresence.ts
@@ -283,6 +283,10 @@ export class LocalPresence implements Presence {
       }
     }
 
+    public setMaxListeners(number: number) {
+      this.channels.setMaxListeners(number);
+    }
+
     public shutdown() {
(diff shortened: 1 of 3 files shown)


endel:
Thank you so much @sylvainpolletvillard, just bumped versions of `@colyseus/core` and `@colyseus/redis-presence` with your addition 🙌

PR #860, opened 2025-05-25
Make presence api always return promises

sylvainpolletvillard:
Currently Presence API are a mix of asynchronous and synchronous calls depending if you use LocalPresence or RedisPresence.

Some functions like `hset` also do not have any typed return signature. This has been a footgun for me because it can lead to race conditions not spotted by the TypeScript compiler:

```
// missing await keyword is not reported as an error
// this works locally with LocalPresence, but fails on production silently with RedisPresence
for (const bot of botsData) {
       matchMaker.presence.hset("bots", bot.id, JSON.stringify(bot))
}
```

I propose to make the Presence API functions signature safer by restricting them to asynchronous functions only, that always return promises. This force the user to use the "await" keyword, and make sure they avoid any asynchronous / race condition trap.

The LocalPresence functions have been changed to return a synchronous value with Promise.resolve. This adds a microtask which is an insignificant performance loss, and ensure the behavior of the logic will remain the same when switching from LocalPresence to RedisPresence

### DIFF
--- a/packages/core/src/presence/LocalPresence.ts
+++ b/packages/core/src/presence/LocalPresence.ts
@@ -12,290 +12,291 @@ type Callback = (...args: any[]) => void;
 const DEVMODE_CACHE_FILE_PATH = path.resolve(".devmode.json");
 
 export class LocalPresence implements Presence {
-    public channels = new EventEmitter();
+  public channels = new EventEmitter();
 
-    public data: {[roomName: string]: string[]} = {};
-    public hash: {[roomName: string]: {[key: string]: string}} = {};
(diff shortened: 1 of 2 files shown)


endel:
Thank you for the contribution @sylvainpolletvillard ❤️  

I've changed a few other places to force using Promises as well:
- `llen`
- `rpush`
- `lpush`
- `lpop`
- `rpop`

sylvainpolletvillard:
Note this is technically a breaking change if some people were relying on the synchronous return values of some functions of LocalPresence. I don't know if they are many, but it might be safer to keep this for 0.17 ? Not sure how you manage versioning for colyseus

PR #918, opened 2026-03-05
Show Process not System Memory

### LINKED ISSUE #853: [Bug]: @colyseus/monitor Memory displays the system memory instead of Node.js process memory
### Context

Image

Image

### Bug description

I created a new colyseus server folder, no changes or adjustments.

I ran "npm start" in the folder and when I went to the local host, I saw the amount of memory displayed fluctuated from 6-12 GB. When I checked with devTool, I didn't see any unusual phenomenon causing such a large amount of Memory consumption?

when i create and dispose room in the first time, i can not create room again from client, The monitor cann't reload too.

### Reproduction

_No response_

### Steps to reproduce

npm create colyseus-app@latest ./uno-server
 
cd uno-server
 
npm start

### Environment & Versions

Colyseus 0.16.3

GABeech:
This change fixes #853

The PR uses the `process.memoryInfo()` call to get memory statistics for the process instead of the whole system.

It also updates the backend API to Send: 

* RSS
* heapUsed
* heapTotal

This allows the frontend to display some more detailed info. 

One thing I'm not totally happy with is the waynthat I am currenly displaying the data. 

image

### DIFF
--- a/packages/monitor/src-backend/api.ts
+++ b/packages/monitor/src-backend/api.ts
@@ -1,94 +1,104 @@
-import { matchMaker } from '@colyseus/core';
+import { matchMaker } from "@colyseus/core";
 
-import express from 'express';
-import { OSUtils } from 'node-os-utils';
+import express from "express";
+import { OSUtils } from "node-os-utils";
 
-import type { MonitorOptions } from './index.js';
+import type { MonitorOptions } from "./index.js";
 
 const osutils = new OSUtils();
 
-const UNAVAILABLE_ROOM_ERROR = "@colyseus/monitor: room $roomId is not available anymore.";
+const UNAVAILABLE_ROOM_ERROR =
+  "@colyseus/monitor: room $roomId is not available anymore.";
 
-export function getAPI (opts: Partial<MonitorOptions>): express.Router {
-    const api = express.Router();
+export function getAPI(opts: Partial<MonitorOptions>): express.Router {
+  const api = express.Router();
 
-    api.get("/", async (req: express.Request, res: express.Response) => {
-        try {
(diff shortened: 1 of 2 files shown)


idirdev (commented):
The switch from system memory (`node-os-utils`) to process memory (`process.memoryUsage()`) makes sense for a game server monitor — operators care about how much memory *their process* is using, not the whole system. A few things to consider:

**1. Breaking API change**

The `/` endpoint's `memory` field changes from `{ totalMemMb, usedMemMb }` (megabytes) to `{ rss, heapTotal, heapUsed }` (bytes). Anyone consuming this API programmatically (custom dashboards, alerting scripts) will break. Worth calling this out in the PR description or bumping the version accordingly.

**2. Dead code**

`formatMemory(memInMb)` (line ~146) is no longer called anywhere since the display now uses `bytesToStr()` directly. Should be removed.

**3. Formatting noise**

The diff is ~420 lines but the actual feature change is ~30 lines. The rest is quote style (`'` → `"`), indentation, trailing commas, and whitespace. Separating formatting into its own commit would make it easier to review the logic changes.

**4. `bytesToStr(0)` guard**

Good catch adding the `size === 0` guard — the original would hit `Math.log(0) = -Infinity` and produce `NaN`. One edge case: negative values (shouldn't happen with `process.memoryUsage()`, but a guard like `if (size <= 0) return "0 B"` would be more defensive).

**5. UI**

You mentioned not being totally happy with the display. The 3 chips side-by-side might get cramped on smaller screens. An alternative could be showing just RSS as the primary chip (it's the "real" memory footprint) and putting heap details in a tooltip or expandable section.

PR #934, opened 2026-04-24
sdk: H3Transport reassemble frames across reads

anaibol:
## Problem

`H3TransportTransport.readIncomingData()` (and its unreliable counterpart) in `packages/sdk/src/transport/H3Transport.ts` assumes every `reader.read()` chunk contains one or more **whole** length-prefixed frames:

```ts
const messages = result.value;
const it: Iterator = { offset: 0 };
do {
    // QUESTION: should we buffer the message in case it's not fully read?
    const length = decode.number(messages as any, it);
    this.events.onmessage({ data: messages.subarray(it.offset, it.offset + length) });
    it.offset += length;
} while (it.offset < messages.length);
```

With WebTransport over HTTP/3, a single stream read is not guaranteed to land on a frame boundary. A chunk can:

1. End **mid-payload** — `onmessage` is called with a truncated buffer and the schema decoder then fails on the next patch with partial-decode errors (stale `refId`, missing fields, "no welcome data" during handshake).
2. End **inside the varint length prefix** — `decode.number` throws and the whole read loop aborts.

In practice this manifests as sporadic handshake failures and `ROOM_STATE_PATCH` decode errors on rooms with larger initial state, especially under dev / local latency where chunks are naturally fragmented.

## Fix

Introduce a `FrameReassembler` that buffers a pending byte tail across reads and only emits fully-received frames. Both the reliable and unreliable readers feed a reassembler instance and dispatch complete frames to `onmessage`.

This answers the existing `// QUESTION: should we buffer the message in case it's not fully read?` TODO — yes, we should.

No change to the wire format, message ordering, or public API of `H3TransportTransport`. `FrameReassembler` is exported so it can be unit-tested directly; it's a narrow helper class.

## Test plan

Unit tests in `packages/sdk/test/h3transport.test.ts` cover:

- [x] single whole frame in one chunk
- [x] multiple whole frames packed in one chunk
- [x] payload split across two reads
- [x] multi-byte varint length prefix split across two reads
- [x] mixed whole + partial frames across reads
- [x] empty / `undefined` chunks (no-op)

```
 RUN  v2.1.9 /private/tmp/colyseus/packages/sdk
 ✓ test/h3transport.test.ts (6 tests) 2ms
 Test Files  1 passed (1)
      Tests  6 passed (6)
```

## Notes

Originally scoped for `colyseus/colyseus.js`, but that repo is archived; active development lives in this monorepo at `packages/sdk`. The bug is present in both.

### DIFF
--- a/packages/sdk/src/transport/H3Transport.ts
+++ b/packages/sdk/src/transport/H3Transport.ts
@@ -1,10 +1,72 @@
-import { encode, decode, type Iterator } from '@colyseus/schema';
-import type { ITransport, ITransportEventMap } from "./ITransport.ts";
+import { ITransport, ITransportEventMap } from "./ITransport";
+import { encode, decode, Iterator } from '@colyseus/schema';
+
+// 9 bytes is the maximum length of a variable-length integer prefix
+const MAX_LENGTH_PREFIX_BYTES = 9;
+
+/**
+ * Reassembles length-prefixed frames from arbitrary byte chunks.
+ *
+ * A single WebTransport `reader.read()` may:
+ *   - deliver multiple whole frames in one chunk
+ *   - split a frame (or its length prefix) across multiple chunks
+ *
+ * This reassembler buffers partial data across reads so each dispatched
+ * frame is exactly one complete message.
+ */
+export class FrameReassembler {
+    private pending: Uint8Array = new Uint8Array(0);
+
+    push(chunk: Uint8Array | undefined): Uint8Array[] {
+        if (!chunk || chunk.byteLength === 0) { return []; }
+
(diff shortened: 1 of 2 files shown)


endel:
Thank you for reporting @anaibol. We had the same issue in the server-side - I've applied the same `FrameReassembler` technique at the server.  Just published new versions of [`@colyseus/sdk`](https://github.com/colyseus/colyseus/releases/tag/%40colyseus%2Fsdk%400.17.42) and [`@colyseus/h3-transport`](https://github.com/colyseus/colyseus/releases/tag/%40colyseus%2Fh3-transport%400.17.11) 🙌

PR #936, opened 2026-05-03
fix(core): emit per-view ref introductions before encodeAll baseline in getFullState (closes #935)

### LINKED ISSUE #935: SchemaSerializer.getFullState produces incomplete snapshot for second filtered client when non-@view fields mutate between joins
# `getFullState` produces a snapshot referencing unregistered refIds for the second filtered client when a non-@view field has accumulated mutations between joins

## Versions

- `@colyseus/schema@4.0.21`
- `@colyseus/core@0.17.42`
- Node + Bun (reproduced under both)

## Symptom

In a room with a `StateView`-filtered `MapSchema` field and a non-`@view` `Schema` field, the second client to join after some non-`@view` state has been mutated receives a snapshot whose first patch references a refId the client decoder has never registered.

Client log:

```text
"refId" not found: 238 { previousRef: ..., previousRefId: 0 }
@colyseus/schema: Please report this issue to the developers.
```

The first client receives a complete snapshot. After that client joins, the room mutates `state.scoreboard.mode = "deathmatch"` (or any field on a non-`@view` nested Schema). A second client then connects and its snapshot references the scoreboard refId without first receiving the structural ADD that introduces that refId to its decoder.

## Wire evidence

Client A's snapshot includes the root structural ADDs:

```text
0e ff 00 80 01 81 02 82 03 83 04 88 05 ...
   ^^^^^ SWITCH root, ADD field 8 = ref 5
```

Client B's snapshot starts by switching directly to the nested ref:

```text
0e ff cc ee 80 aa "deathmatch" ...
   ^^^^^^^^ SWITCH ref 238 directly, without prior op introducing 238
```

## Root cause

`SchemaSerializer.getFullState` caches `encoder.encodeAll` output and re-runs it when `root.changes.next !== undefined`:

```ts
if (this.needFullEncode || this.encoder.root.changes.next !== undefined) {
  this.sharedOffsetCache = { offset: 1 };
  this.fullEncodeCache = this.encoder.encodeAll(this.sharedOffsetCache, this.fullEncodeBuffer);
  this.needFullEncode = false;
}
```

For client B, `needFullEncode` has already been cleared by client A, but `root.changes.next` is non-empty because of the mid-game mutation. `encodeAll` re-runs, but the structural ADD operations emitted during the earlier full-state pass are no longer queued. The resulting baseline contains UPDATE ops that reference nested refIds without introducing them to this late client.

## Reproduction

A minimal reproduction has been added in PR #936:

- `RoomWithFilteredAndPublic`: one `@view`-filtered `MapSchema` plus one non-`@view` nested `Schema`.
- Test flow: client A joins, non-`@view` nested state mutates, client B joins, assert B decodes without `"refId not found"` and sees the latest nested state.

## Proposed fix

PR #936 reorders the filtered full-state response so pending per-view introductions are encoded before the cached `encodeAll` baseline:

```text
[protocol byte][view introductions][switch root][encodeAll baseline][per-view filtered ops]
```

This ensures a late client's decoder sees the structural ref introductions before any baseline updates reference those refs.

## Downstream validation

In the downstream Abraxas codebase, I verified the patched serializer through focused serializer/AOI/room-sync suites and re-ran the local duel-bot harness (`scripts/client/duelBots.ts`). Two 60s two-bot runs joined both clients with:

```text
invalid byte = 0
field not defined = 0
fatal = 0
```

There are still a few app-specific AOI `refId not found` misses in that larger harness, so I am not claiming this issue/PR fixes every possible StateView/AOI edge case. This issue is specifically about the late-joiner snapshot missing the non-`@view` parent ref introduction.

## Notes

- `@colyseus/schema@4.0.21` already includes #218. That fix is related to sub-schema visibility under filtered parents, but this issue is different: the parent field's structural `set` op is missing entirely.
- Rooms with only filtered fields, or only non-filtered fields, may not trigger this. The reproduction needs both a filtered view path and a non-`@view` nested Schema mutation between joins.

anaibol:
## Summary

Fixes #935: a filtered client that calls `getFullState()` after non-`@view` state has been mutated can receive a snapshot whose first patch references refIds that the client decoder has never registered.

## Root Cause

`SchemaSerializer.getFullState()` caches `encoder.encodeAll()` output. For the first filtered client, `encodeAll()` can still walk the root `allChanges` list and emit the structural `ADD` operations that introduce non-`@view` child refs.

After a later mutation invalidates the full-state cache via `root.changes.next !== undefined`, `encodeAll()` runs again, but the earlier structural introductions may no longer be present in the linked list. The resulting baseline can switch directly to a nested ref and emit updates for it before the late client's decoder has received the operation that introduces that ref from the root.

In wire-order terms, the broken shape is:

```text
[protocol byte][baseline that switches to nested ref before introducing it][per-view filtered ops]
```

## Fix

When a client's `view.changes` has pending entries, encode those per-view introductions first and splice them between the protocol byte and the `encodeAll()` baseline. Because those introductions may leave the decoder positioned on a non-root ref, the patched output switches back to the root ref before replaying the baseline.

The corrected wire order is:

```text
[protocol byte][view introductions][switch root][encodeAll baseline][per-view filtered ops]
```

The existing `view.changes.size === 0` path stays a no-op and preserves the previous behavior for clients that have no pending per-view introductions.

## Test Plan

- Added `RoomWithFilteredAndPublic`, a regression fixture with one `@view`-filtered `MapSchema` and one non-`@view` nested `Schema`.
- Added a `packages/testing/test/Usage.test.ts` case covering: client A joins, non-`@view` nested state mutates, client B joins, B decodes without `"refId" not found`, and B sees the latest nested state.
- CI is passing on this PR.

## Notes

This is a `getFullState()` ordering fix for late filtered snapshots. It does not change the unfiltered full-state path, and it does not change the `view.changes.size === 0` fast path.

### DIFF
--- a/bundles/colyseus/package.json
+++ b/bundles/colyseus/package.json
@@ -20,6 +20,7 @@
     }
   },
   "scripts": {
+    "pretest": "pnpm --filter @colyseus/monitor --filter @colyseus/playground run build",
     "test": "COLYSEUS_SEAT_RESERVATION_TIME=0.3 COLYSEUS_PRESENCE_SHORT_TIMEOUT=300 mocha test/**.test.ts test/**/**.test.ts --exit --timeout 15000"
   },
   "author": "Endel Dreyer",
--- a/bundles/colyseus/test/ExceptionHandling.test.ts
+++ b/bundles/colyseus/test/ExceptionHandling.test.ts
@@ -8,10 +8,8 @@ const TEST_PORT = 8570;
 const TEST_ENDPOINT = `ws://localhost:${TEST_PORT}`;
 
 async function waitFor(predicate: () => boolean, timeoutMs: number = 500) {
-  const startedAt = Date.now();
-  while (!predicate() && Date.now() - startedAt < timeoutMs) {
-    await timeout(10);
-  }
+  const deadline = Date.now() + timeoutMs;
+  while (!predicate() && Date.now() < deadline) await timeout(10);
 }
 
 describe("Exception Handling", () => {
--- a/bundles/colyseus/test/ExceptionHandling.test.ts
+++ b/bundles/colyseus/test/ExceptionHandling.test.ts
@@ -7,6 +7,13 @@ import { timeout } from "./utils/index.ts";
 const TEST_PORT = 8570;
 const TEST_ENDPOINT = `ws://localhost:${TEST_PORT}`;
 
+async function waitFor(predicate: () => boolean, timeoutMs: number = 500) {
+  const startedAt = Date.now();
+  while (!predicate() && Date.now() - startedAt < timeoutMs) {
+    await timeout(10);
+  }
+}
+
 describe("Exception Handling", () => {
   let server: Server;
   let client = new Colyseus.Client(TEST_ENDPOINT);
@@ -351,7 +358,7 @@ describe("Exception Handling", () => {
     });
 
     await client.joinOrCreate("my_room", { arg0: "arg0" });
-    await timeout(110);
+    await waitFor(() => caught.error !== undefined);
 
     assert.ok(caught.error instanceof TimedEventException);
     assert.strictEqual(caught.error.message, "setTimeout Error");
@@ -436,7 +443,7 @@ describe("Exception Handling", () => {
 
     const conn = await client.joinOrCreate("my_room", { arg0: "arg0" });
     await conn.send("foo", "bar");
-    await timeout(50);
+    await waitFor(() => caught.error !== undefined);
     await conn.leave();
 
     assert.ok(caught.error instanceof OnMessageException);
--- a/bundles/colyseus/test/Presence.test.ts
+++ b/bundles/colyseus/test/Presence.test.ts
@@ -325,8 +325,7 @@ describe("Presence", () => {
 
       describe("brpop", () => {
         beforeEach(async () => {
-          await presence.del("brpop");
-          await presence.del("none");
+          await Promise.all([presence.del("brpop"), presence.del("none")]);
         });
 
         it("brpop should return existing item", async () => {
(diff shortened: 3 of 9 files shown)


anaibol:
Added a regression test in `4aabd7db`:

- New fixture `RoomWithFilteredAndPublic` (`packages/testing/test/app1/RoomWithFilteredAndPublic.ts`): minimal room with one `@view`-filtered `MapSchema` and one non-`@view` nested `Schema`.
- New test case in `packages/testing/test/Usage.test.ts` (`describe("late-joiner snapshot (#935)")`) that:
  1. Has client A join.
  2. Mutates the non-`@view` nested between joins. This is the trigger: it invalidates `getFullState`'s `encodeAll` cache via `root.changes.next`.
  3. Has client B join.
  4. Asserts no `"refId not found"` errors during B's snapshot decode via `sinon.spy` on `console.error`.
  5. Asserts B sees the latest state of the nested field.

endel:
Hi @anaibol, thanks for the report and proposed solution. The fix for this specific issue should live in the `@colyseus/schema` repository. It's very likely related to encoding order of `view.changes`. We had a similar issue reported on Discord [here](https://discord.com/channels/525739117951320081/1499152544520343582/1499425378849001612).

I'm working on a fix locally which I hope to release later today 🙏

anaibol:
> Hi @anaibol, thanks for the report and proposed solution. The fix for this specific issue should live in the `@colyseus/schema` repository. It's very likely related to encoding order of `view.changes`. We had a similar issue reported on Discord [here](https://discord.com/channels/525739117951320081/1499152544520343582/1499425378849001612).
> 
> I'm working on a fix locally which I hope to release later today 🙏

nice! thanks! will check that update later and create a PR on the schema repo next time.

endel:
Hi @anaibol, [version `4.0.22` of `@colyseus/schema` is out](https://github.com/colyseus/schema/releases/tag/4.0.22), please let me know if you still have any issues using it!

anaibol:
Hi @endel, thanks for the `4.0.22` release. I checked it against our local regression coverage.

The `StateView` topological ordering fix does address the ordering class from this PR, but while validating the upgrade I still found two schema-level edge cases around view snapshots/patches:

- `Encoder.encodeAllView()` does not consume pending manual `StateView.add()` / `view.changes`, while `encodeView()` does.
- stale invalid `view.changes` entries can still throw before the encoder skips them.

I opened the follow-up in the schema repo here: colyseus/schema#225.

PR #947, opened 2026-07-23
fix(core): support strict TypeScript declaration consumers

Hoodgail:
## Summary

- preserve `Room` lifecycle override inference while handling optional hooks in exception declarations
- prevent `LocalPresence.subscriptions` from leaking an invalid inferred `EventEmitter<[never]>` type
- constrain the functional `room()` helper generic to `RoomOptions`
- add a strict NodeNext declaration regression covering the generated package output

## Verification

- `pnpm run build`
- `pnpm --filter @colyseus/core test`
- packed `@colyseus/core` and the schema fix from colyseus/schema#227 into a fresh TypeScript 5.9.3 NodeNext consumer with `strict: true` and `skipLibCheck: false`
- runtime import smoke test for `Room` and `LocalPresence`

The recursive workspace test command remains blocked in an unrelated workspace because its `vitest` executable is unavailable; the root build, focused core test, packed declaration consumer, and runtime smoke test pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -3,6 +3,9 @@
   "version": "0.17.44",
   "description": "Multiplayer Framework for Node.js.",
   "type": "module",
+  "scripts": {
+    "test": "tsc --noEmit --strict --skipLibCheck false --module NodeNext --moduleResolution NodeNext --target ES2022 --typeRoots ../../node_modules/@types --types node test/strict-declarations.ts && node --input-type=module -e \"import assert from 'node:assert/strict'; import fs from 'node:fs'; assert.match(fs.readFileSync('build/presence/LocalPresence.d.ts', 'utf8'), /subscriptions: EventEmitter;/);\""
+  },
   "input": "./src/index.ts",
   "main": "./build/index.cjs",
   "module": "./src/index.ts",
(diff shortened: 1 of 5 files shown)


endel:
Thank you @Hoodgail, great catch — and much appreciated the thorough verification steps in the description! 🙌

This is now published in [`@colyseus/core@0.17.45`](https://github.com/colyseus/colyseus/blob/master/packages/core/CHANGELOG.md#01745).

A couple of follow-up tweaks I made on top of your work in f8a0958:

- Moved the compiler flags from the inline `test` script into `test/tsconfig.json`, so the script is now just `tsc -p test`.
- Replaced the regex assertion on `LocalPresence.d.ts` with a type-level check (`Equal<LocalPresence['subscriptions'], EventEmitter>`) — same regression coverage, but resolved through the package `exports` the way a real consumer sees it, and immune to emit formatting changes.
- Dropped the `room()`-specific assertions from the test file: the functional `room()` helper is deprecated and already removed in 0.18, so keeping them would cause conflicts when merging into the 0.18 branch. Your `T extends RoomOptions` constraint is kept as-is — with `skipLibCheck: false` the strict compile still validates the `room()` declaration on 0.17 either way.

Thanks again for the contribution! 🚀

PR #953, opened 2026-08-10
allowReconnection and findOneRoomAvailable document arguments they do not take

darkdi:
Was tracing the reconnection flow and my editor showed `allowReconnection` taking a `client`, which it does not, the argument is `previousClient`. The same block ships in `@colyseus/core@0.17.49` under `build/Room.d.ts`, so that tooltip is wrong wherever anyone hovers it.

`findOneRoomAvailable` is the messier one. The block says `sortOptions`, the argument is `additionalSortOptions`, and there is a local `sortOptions` a few lines down holding the handler defaults with the argument merged over them. So the documented name points at a different value than the one you pass in. I reworded that description line instead of only renaming it, since a bare rename still reads as if it were the final sort order.

Raising this against your note about cosmetic patches: both names reach users through editor tooltips and the generated docs, and one of them collides with a different value inside the same function.

### DIFF
--- a/packages/core/src/MatchMaker.ts
+++ b/packages/core/src/MatchMaker.ts
@@ -326,7 +326,7 @@ export async function query<T extends Room = any>(
  *
  * @param roomName - The Id of the specific room.
  * @param filterOptions - Filter options.
- * @param sortOptions - Sorting options.
+ * @param additionalSortOptions - Sorting options, merged over the ones declared on the room handler.
  *
  * @returns Promise<IRoomCache> - A promise contaning an object which includes room metadata and configurations.
  */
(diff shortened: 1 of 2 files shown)


endel:
Merged — thank you.

You were right that this isn't merely cosmetic, and right about which one is worse: `findOneRoomAvailable`'s documented `sortOptions` names a *different value* than the argument, so the tooltip actively misleads. Your rewording is what shipped, for the reason you gave — a bare rename still reads as if the argument were the final sort order.

One correction to my earlier comment: I'd applied the change by hand while this PR sat open, so it went out in **0.17.50** (`latest`) and **0.18.4** (`next`) before the merge. That was the wrong order — I've reverted the manual commit on `master` so your commits are what actually carry the change, on the lines themselves. You're credited in both changelogs.

PR #963, opened 2026-09-08
fix(redis-presence): ignore malformed subscription messages

### LINKED ISSUE #730: [Bug]: Calling `ioredis.client.publish("SOME_TOPIC", "")` will crash Colyseus (game) server
### Context

We have another server connected to the same Redis as the game server. The other server is using `ioredis` npm package for the connection. And it will publish (i.e. "SOME_TOPIC") a change, then the game server (subscribed to the same topic) will react.

### Bug description

If an empty string is passed in as the payload. For example:
ioredis.client.publish("SOME_TOPIC", `""`) 
This would crash the game server since it expect a JSON
A workaround is to pass in:
ioredis.client.publish("SOME_TOPIC", `"{}"`) 

### Reproduction

_No response_

### Steps to reproduce

1. run colyseus game server with RedisPresence
2. game server Presence subscribe to "SOME_TOPIC"
3. run another script/server, connect to redis using ioredis, and fire a `.publish("SOME_TOPIC", "")`
4. game server crash

### Expected

From the error message, it seems that there is a `JSON.parse("")` in the `RedisPresence.handleSubscription` function, it would be nice if it is handled.

### Environment & Versions

@colyseus/core": "0.15.27"
Node: 18
"typescript": "4.9.5"

jokrasno:
An external Redis publisher can send an empty or malformed JSON payload to a subscribed channel, causing `RedisPresence` to throw out of its message event handler. Catch JSON parsing errors, warn with the channel name, and discard that message so later valid messages can still be delivered. Application callback errors remain outside the catch.

Closes https://github.com/colyseus/colyseus/issues/730

Validation:
- Added five focused tests using the ioredis message event with lazy connections; no Redis server is required. The three malformed-input cases failed on the original implementation; all five tests pass with the fix (including valid falsy JSON values and callback errors).
- Ran `pnpm build --scope @colyseus/redis-presence`, then `node --import tsx node_modules/mocha/bin/mocha.js bundles/colyseus/test/presence/RedisPresence.messages.test.ts --timeout 5000`.
- The scoped build emitted JS and declarations, but reported an existing `void | Response` diagnostic in `@colyseus/core/src/Transport.ts:107` when using the published `@colyseus/core@next` dependency in the isolated test setup. The full monorepo/Redis integration suite was not run.

---

Maintainer follow-up (9d20f8f):
- Changelog entry moved under a real `0.18.4` heading (the repo doesn't use `Unreleased`), with the package version bumped in the same commit.
- The five tests folded into the existing `bundles/colyseus/test/presence/RedisPresence.test.ts` as a `message handling` block, rather than a second file.
- Verified against a live Redis: the three malformed cases fail on the original implementation and the full presence suite passes (66) with the fix. No new `tsc` errors.


endel:
Thank you for the PR @jokrasno 👏

PR #967, opened 2026-09-16
Fix Vite Plugin Module Loading Behavior 

MeirArani:
The current implementation of the colyseus vite plugin doesn't play well other vite plugins. In my case, `nitro`.

Other plugins might cause the colyseus environment's module runner to be inert (i.e., `undefined` and unable to perform `import`). This is because the environment shifts from a `RunnableDevEnvironment` (which allows for `runner.import`) into a `FetchableDevEnvironment` (which has no built in runner). This stops the plugin from loading the server's `app.config.ts` file. 

While digging around for advice on the issue, I discovered the Cloudflare vinext package was [facing the same issue](https://github.com/cloudflare/vinext/issues/853). They solved the problem by building a simple `ModuleRunner` named `DirectRunner` that works with `FetchableDevEnvironment`. From there, the solution is just as simple as running:

```ts
env.runner = createDirectRunner(env);
env.runner.import(blahblahblah);
env.runner.close();
```

The cloudflare code is also under MIT and well documented, so I opted to copy it over (with attribution).

[Vite's own documentation](https://vite.dev/guide/api-environment-frameworks) also advises using `FetchableDevEnvironment` over `RunnableDevEnvironment`:

> FetchableDevEnvironment is an environment that can communicate with its runtime via the [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Window/fetch) interface. Since the RunnableDevEnvironment is only possible to implement in a limited set of runtimes, we recommend to use the FetchableDevEnvironment instead of the RunnableDevEnvironment.

The specifics of enforcing a `FetchableDevEnvironment` in the colyseus vite plugin are probably beyond the scope of this PR, but I wanted to present this as a starting point for this issue.


endel:
Thanks for the PR and the details @MeirArani, it would be super to support Nitro, I'm afraid we may need more specific changes in order to work on all deployment targets supported by Nitro (CloudFlare, Vercel, AWS Lambda, etc) - but this is a good start!

Just published [`colyseus@0.18.6`](https://github.com/colyseus/colyseus/releases/tag/colyseus%400.18.6) with the changes applied here! Gonna close this but feel free to open another issue/PR as follow-up if needed!

🤖🥩🫳 meat proxying below...

---
Your original PR had a few problems:

- **It breaks every project without Nitro.** `RunnableDevEnvironment.runner` only has a getter, so `env.runner = ...` throws. The plugin catches the error and starts with no rooms:
  ```
  TypeError: Cannot set property runner of #<RunnableDevEnvironment> which has only a getter
  ```
- **It closes the runner after each import.** A closed runner can't import again, so dynamic imports in your server code fail with `Vite module runner has been closed.`
- **`createNodeImportMeta` needs Vite 7.1.** The plugin supports Vite 6 and later. On Vite 6 and 7.0, `colyseus/vite` fails to load.

The fixes, based on your approach:

- The plugin keeps Vite's runner when the environment has one. Otherwise, it creates a runner on `fetchModule()`, once per environment, and reuses it across reloads.
- The trigger is Nitro itself. Nitro adopts any server environment with a string or array build input as one of its services. The plugin now declares its input as an object, so Nitro doesn't adopt the environment.
- The same change fixes the production build. Before, Nitro moved the server bundle into `node_modules/.nitro/`, and `dist/server/server.mjs` was missing.
- A regression test covers an environment without a runner, including a reload.

I verified the fix on a TanStack Start app with `nitro()`: development, `vite build`, and production, with an SDK client joining a room.

PR #969, opened 2026-09-19
fix(monitor): display process RSS instead of system memory

### LINKED ISSUE #853: [Bug]: @colyseus/monitor Memory displays the system memory instead of Node.js process memory

fatihcvs:
The monitor currently labels host-wide memory usage as server memory. Display the serving Node.js process's resident memory (RSS) instead, via a new `memory.rssMb` field. Existing `totalMemMb` and `usedMemMb` fields retain their meanings and units.

Fixes #853. This follows the invitation to contribute on that issue; #918 was closed without merging. This patch keeps the API compatible and the UI change small. In a multi-process deployment, this is the process serving the monitor, not the sum of all workers.

| Before | After |
| --- | --- |
| Host memory | Process RSS |

Validation on Windows, Node 24.13.0 / pnpm 10.9.0:

- New endpoint regression test fails before the fix and passes after; checks byte-to-MB conversion, refreshed values on the next request, and retained system memory fields.
- 11 existing dual-mode middleware tests pass.
- Root and monitor builds exit successfully. They report existing declaration diagnostics; monitor `tsc --noEmit` produces identical errors before/after (RoomInspect, uninitialized interval, and CSS import typing).
- Local browser/API comparison at 1280px and 390px: host memory around 21 GB, process RSS around 78 MB, matching the serving process. The existing narrow-screen horizontal overflow remains. The full Redis-dependent monorepo suite was not run.

AI-assisted implementation and validation.


endel:
Thank you for another contribution @fatihcvs 🥳

fatihcvs:
You're welcome, @endel! Happy to contribute.

PR #970, opened 2026-09-19
fix(monitor): prevent page overflow on small screens

fatihcvs:
At phone widths, the monitor's statistics force the page wider than the viewport; the room title and action buttons also compete for the same row. Stack the cards below `sm` and put room actions on a separate row. The existing layout at `sm` and above is preserved.

Verified against a real local Express monitor and in-memory schema room:

- Before: list scroll width 452px at a 320px viewport; room detail 376px. After: no page overflow at 320, 390, 599, 600 or 1280px, for empty/populated lists and room detail. Desktop stat-card positions are unchanged.
- Inspect, State tab, Broadcast dialog open/cancel and Back navigation work. Data grids retain their own horizontal scrolling.
- Root `pnpm build` and monitor build exit 0 with existing declaration diagnostics. Monitor `tsc --noEmit` reports exactly the same pre-existing diagnostics as the upstream baseline; it is not a clean type check. No full Redis-dependent suite was run for this styling change.

[Browser evidence and measurements](https://github.com/fatihcvs/colyseus/tree/docs/monitor-mobile-evidence/visual-evidence/monitor-mobile)

| 390px | Before | After |
| --- | --- | --- |
| Room list | Before | After |
| Room detail | Before | After |


endel:
Good stuff, thank you for the PR @fatihcvs 🥳

fatihcvs:
You're welcome, @endel! Glad to help improve the monitor on mobile.

PR #971, opened 2026-09-20
fix(monitor): restore TypeScript checking

fatihcvs:
The monitor's `tsc --noEmit` currently fails on five diagnostics: uninitialized polling timer fields, generic grid IDs passed to string session handlers, and the CSS side-effect import. This restores a clean check and adds `pnpm --filter @colyseus/monitor typecheck` to make it easy to run.

- Model timer handles as possibly undefined before polling starts.
- Type client columns from the client state and use their string sessionId for Send/Disconnect; row IDs already contain that same value.
- Use the DataGrid formatter signature instead of `Function`, removing the elapsed-time column assertion.
- Include Vite's client declarations for CSS imports. No compiler strictness settings are relaxed.

Validation on Node 24.13.0 / Windows, based on master 038b40c:
- Before: `pnpm --filter @colyseus/monitor exec tsc --noEmit` fails with five diagnostics. After: the new `typecheck` script passes with no diagnostics.
- Root `pnpm build` and monitor build exit 0. Existing backend declaration diagnostics remain; these are separate from the clean tsconfig-based check.
- Both existing Monitor API tests pass against the rebuilt package (Mocha via `node --import tsx`, the Windows equivalent of the repository test runner). The full Redis-dependent suite was not run for this typing change.

CI note: the PR's packaging job passes. The test-suite job reaches `packages/example` and fails in the admin e2e setup because Puppeteer's Chrome 150.0.7871.24 is missing. The [base master run at 038b40c](https://github.com/colyseus/colyseus/actions/runs/35486001069) fails with the identical missing-Chrome error; [this PR run](https://github.com/colyseus/colyseus/actions/runs/35493018923) does too. The failure is outside the monitor typing changes.


endel:
Thank you @fatihcvs

fatihcvs:
You're welcome, @endel! Happy to help keep the monitor's type checks working.

PR #972, opened 2026-09-20
fix(ci): install Chrome for admin e2e tests

fatihcvs:
The test-suite job fails during the admin e2e setup because Puppeteer's expected Chrome binary is missing (`Could not find Chrome (ver. 150.0.7871.24)`). This is also present on [master at 038b40c](https://github.com/colyseus/colyseus/actions/runs/35486001069); 40 admin tests are cancelled before running.

Explicitly install Chrome through the example package's Puppeteer CLI after dependency installation. The installed Puppeteer version selects its compatible browser revision, so this does not hard-code a Chrome version or depend on an install lifecycle hook having populated the browser cache. Only the test-suite job needs the browser.

Validation:
- With a new empty browser cache, the existing Puppeteer launch fails with the same missing-Chrome error.
- `pnpm --filter @colyseus/example exec puppeteer browsers install chrome` installs the expected binary; a subsequent headless launch reports `Chrome/150.0.7871.24` and exits cleanly.
- The smoke test ran on Windows/Node 24, with all downloaded files on D:. The [PR workflow on Ubuntu](https://github.com/colyseus/colyseus/actions/runs/35493623220) now passes: Chrome installation, build-all, the full test:ci suite (including admin e2e), and the separate packaging job.
- `git diff --check` passes. No child package or dependency changes.


endel:
Great stuff, thank you again @fatihcvs 🙏

fatihcvs:
You're welcome! Thanks for reviewing and merging the CI fix.

PR #973, opened 2026-09-20
fix(geoip): retry failed reader initialization

fatihcvs:
A failed GeoIP database load remains in the shared reader cache as a rejected promise. Later rooms reuse that rejection even after the database file becomes available, so room creation cannot recover without restarting the process.

Evict the failed promise only if it is still the current cache entry, then rethrow the original error. Successful readers remain shared between rooms; the identity check also avoids deleting a newer cache entry.

Validation uses the existing real MMDB fixture and Room/plugin lifecycle:
- Start with a missing database and verify the original ENOENT rejection.
- Copy the valid fixture into place and create another room. This failed with the cached ENOENT before the change.
- After successful loading, remove only the test's temporary copy and confirm another room still resolves the fixture's country from the successful cached reader.

No MaxMind account, download, or external service is used. Root `pnpm build` exits 0 with the existing declaration diagnostics. All 14 GeoIP package tests pass (Mocha with `node --import tsx` on Node 24/Windows); `git diff --check` passes.

The base branch currently has a known missing-Chrome failure in the example admin e2e setup; #972 addresses it and its full Ubuntu suite is green. This change is independent of that CI fix.

CI result: [run 35494249527](https://github.com/colyseus/colyseus/actions/runs/35494249527) passes packaging and all 14 GeoIP tests, including the new recovery regression. The full test-suite reaches the example admin e2e setup and fails on the same missing Chrome 150.0.7871.24 as master; 40 admin tests are cancelled. The independent #972 fix has a green full-suite run.


endel:
Thank you for the contribution @fatihcvs 🙏

fatihcvs:
You're welcome! Thanks for reviewing the GeoIP fix.

